PR #31, opened 2022-03-26
Add access token validator support

mikenachbaur-okta:


mikenachbaur-okta:
Tests are failing, but I've gotta go pick my daughter up from her ski camp, so I'll get them passing on Linux on Monday.

Sources/AuthFoundation/JWT/Internal/DefaultAccessTokenValidator.swift
```diff
@@ -0,0 +1,49 @@
...
+
+struct DefaultAccessTokenValidator: AccessTokenValidator {
+    func validate(accessToken: String, idToken: JWT) throws {
+        #if !canImport(CommonCrypto)
+        throw JWTError.signatureVerificationUnavailable
+        #endif
+
+        guard let atHash = idToken.value(String.self, for: "at_hash")
+        else {
+            return
```
jaynewstrom-okta, line 27: Can you add a test for this case? If the `at_hash` is missing from the ID Token, it should be valid. (This happens for other IdPs like Auth0.

Sources/AuthFoundation/Token Management/AccessTokenValidator.swift
```diff
@@ -0,0 +1,17 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//
+// See the License for the specific language governing permissions and limitations under the License.
+//
+
+import Foundation
+
+public protocol AccessTokenValidator {
```
jaynewstrom-okta, line 15: Can you add docs for this protocol?

jaynewstrom-okta (approved):
Please consider adding an integration test for this flow, to make sure this code path is called when tokens are minted.

PR #33, opened 2022-04-04
Add unique ID for tokens, and associated metadata

mikenachbaur-okta:
This permits tokens to be uniquely identified by UUIDs, and to associate custom developer-assigned metadata to retrieve tokens based on their arguments.

### DIFF
--- a/Sources/AuthFoundation/Token Management/Internal/KeychainTokenStorage.swift
+++ b/Sources/AuthFoundation/Token Management/Internal/KeychainTokenStorage.swift
@@ -127,7 +127,7 @@ class KeychainTokenStorage: TokenStorage {
         delegate?.token(storage: self, removed: id)
 
         if defaultTokenID == id {
-            try setDefaultTokenID(id)
+            try setDefaultTokenID(nil)
         }
     }
     
--- a/Sources/AuthFoundation/Token Management/Internal/KeychainTokenStorage.swift
+++ b/Sources/AuthFoundation/Token Management/Internal/KeychainTokenStorage.swift
@@ -21,40 +21,37 @@ class KeychainTokenStorage: TokenStorage {
 
     weak var delegate: TokenStorageDelegate?
     
-    private(set) lazy var defaultTokenID: UUID? = {
+    private(set) lazy var defaultTokenID: String? = {
         guard let defaultResult = try? Keychain
                 .Search(account: KeychainTokenStorage.defaultTokenName)
                 .get(),
-              let uuidString = String(data: defaultResult.value, encoding: .utf8),
-              let uuid = UUID(uuidString: uuidString)
+              let id = String(data: defaultResult.value, encoding: .utf8)
         else {
             return nil
         }
         
-        return uuid
+        return id
     }()
 
-    func setDefaultTokenID(_ id: UUID?) throws {
+    func setDefaultTokenID(_ id: String?) throws {
         guard defaultTokenID != id else { return }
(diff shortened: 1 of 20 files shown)


Sources/AuthFoundation/Token Management/Internal/KeychainTokenStorage.swift
```diff
@@ -15,71 +15,47 @@
...
 
     weak var delegate: TokenStorageDelegate?
     
-    private lazy var _defaultTokenId: String? = {
-        if let defaultResult = try? Keychain
-            .Search(account: KeychainTokenStorage.defaultTokenName)
-            .get()
-        {
-            return String(data: defaultResult.value, encoding: .utf8)
+    private(set) lazy var defaultTokenID: UUID? = {
```
stevelind-okta, line 24: Might be simpler/more robust to make  it a `String`? UUID is fine of course but can't think of many benefits and you need to keep converting back and forth given that keychain can't store this type.

Sources/AuthFoundation/Token Management/Internal/UserDefaultsTokenStorage.swift
```diff
@@ -24,119 +25,121 @@ class UserDefaultsTokenStorage: TokenStorage {
...
+    private lazy var allTokens: [UUID:Token] = {
+        if let data = userDefaults.data(forKey: UserDefaultsKeys.allTokensKey),
+           let result = try? JSONDecoder().decode([UUID:Token].self, from: data)
+        {
+            return result
+        }
+        return [:]
+    }()
+    
+    private lazy var metadata: [UUID:[String:String]] = {
```
stevelind-okta, line 47: nit style guide 
```suggestion
    private lazy var metadata: [UUID: [String: String]] = {
```

Sources/AuthFoundation/Token Management/Internal/UserDefaultsTokenStorage.swift
```diff
@@ -24,119 +25,121 @@ class UserDefaultsTokenStorage: TokenStorage {
...
     
-    func remove(token: Token) throws {
-        guard let index = allTokens.firstIndex(of: token) else { return }
-        allTokens.remove(at: index)
-        
-        if _defaultToken == token {
-            _defaultToken = nil
-            delegate?.token(storage: self, defaultChanged: nil)
-        }
+    func assign(metadata: [String:String], for id: UUID) throws {
```
stevelind-okta, line 126: For me had to read this before I understood what it meant
```suggestion
    func assignMetadata(_ metadata: [String:String], for id: UUID) throws {
```

stevelind-okta (commented):
Looks good. Couple suggestions:

1. For me the native `UUID` is adding a lot of code and will continue to do so, would prefer `String` type as will be represented if sent to backend etc. Interested to hear your thoughts on the benefits 😄 

2. Style guide on dictionaries `[KeyType: ValueType]`, add a swift linter to this proj so we don't diverge too much? Can use `okta-devices-swift` as a good linter starting point?

Sources/AuthFoundation/Token Management/Internal/UserDefaultsTokenStorage.swift
mikenachbaur-okta: Or, I should go back to my original `setMetadata` name...

laura-rodriguez (approved):

laurarodriguez-okta (approved):

PR #34, opened 2022-04-06
Add ID token claims to token metadata, with methods to filter credentials by claims

mikenachbaur-okta:


Sources/AuthFoundation/Token Management/Token+Context.swift
```diff
@@ -0,0 +1,42 @@
...
+// See the License for the specific language governing permissions and limitations under the License.
+//
+
+import Foundation
+
+extension Token {
+    /// Summarizes the context in which a token is valid.
+    ///
+    /// This includes information such as the client configuration or settings required for token refresh.
+    public struct Context: Codable, Equatable, Hashable {
```
mikenachbaur-okta, line 19: This edit is just to move the extension to its own file.

Sources/AuthFoundation/Token Management/Token+Enums.swift
```diff
@@ -0,0 +1,42 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//
+// See the License for the specific language governing permissions and limitations under the License.
+//
+
+import Foundation
+
+public extension Token {
```
mikenachbaur-okta, line 15: This edit is just to move the extension to its own file.

README.md
```diff
@@ -149,6 +149,63 @@ func signIn() async {
...
+let token = try await flow.resume(with: context)
+```
+
+### Authentication using Native SSO flow
+
+When using the `device_sso` scope, your application can receive a "device secret", which can be used in combination with your user's ID token to exchange new credentials. To use this within your application, you would use the `TokenExchangeFlow` to exchange those sets of tokens.
+
+```swift
+let flow = TokenExchangeFlow(
+    issuer: URL(string: "https://example.okta.com")!,
```
stevelind-okta, line 187: Might be tough to work around, but I prefer to not give examples of code we wouldn't write ourselves (looking at the `!`). Not a dealbreaker for sure because using an if let might be annoying to use. Add a comment disclaimer? 😂

README.md
```diff
@@ -149,6 +149,63 @@ func signIn() async {
...
+
+### Authentication with Username/Password
+
+For simple authentication use-cases, you can use the `ResourceOwnerFlow` class to authenticate with a plain username and password.
+
+```swift
+let flow = ResourceOwnerFlow(issuer: URL(string: "https://example.okta.com")!,
+                             clientId: "abc123client",
+                             scopes: "openid offline_access email profile")
+let token = try await flow.resume(username: "jane.doe", password: "secretPassword")
```
stevelind-okta, line 206: Just so I understand, is the first call `resume` rather than `start`? I wonder if it would be better to have a `start` call that works only when the flow hasn't been started and a `resume` that theoretically works 'whenever'. Tough to say which is less confusing.

Sources/AuthFoundation/AuthFoundation.docc/ManagingUserCredentials.md
```diff
@@ -49,7 +49,45 @@ NotificationCenter.default.addObserver(forName: .defaultCredentialChanged,
...
+    // Use the credential
+}
+```
+
+#### Finding credentials by ID Token claims
+
+If a token contains a valid ``Token/idToken``, the claims it represents are available within the ``Credential/find(where:)`` expression. The object returned within that expression supports the ``HasClaims`` protocol, meaning its values can be queried within that filter.
+
+```swift
+let userCredentials = try Credential.find { metadata in
```
stevelind-okta, line 87: does this throw if not found? I'm missing a bit of context around `Credential` and why we are using class functions here instead of instance.

Sources/AuthFoundation/JWT/ClaimContainer.swift
```diff
@@ -16,13 +16,13 @@ import Foundation
 ///
 /// > Note: This does not apply to JWT, which while it contains claims, it has a different format which includes headers and signatures.
 public protocol ClaimContainer: JSONDecodable {
-    var rawValue: [String:Any] { get }
+    var payload: [String:Any] { get }
```
stevelind-okta, line 19:
```suggestion
    var payload: [String: Any] { get }
```

Sources/AuthFoundation/JWT/ClaimContainer.swift
```diff
@@ -16,13 +16,13 @@ import Foundation
...
 }
 
 extension ClaimContainer {
```
stevelind-okta, line 22:
```suggestion
extension ClaimsContainer {
```
Seems that this is a container with the whole claims payload. If I'm misunderstanding, then disregard 😄

Sources/AuthFoundation/Responses/TokenInfo.swift
```diff
@@ -17,16 +17,11 @@ import Foundation
 /// This provides a convenience mechanism for accessing information related to a token. It supports the ``HasClaims`` protocol, to simplify common operations against introspected information, and to provide consistency with the ``JWT`` class.
 ///
 /// For more information about the members to use, please refer to ``ClaimContainer``.
-public struct TokenInfo: RawRepresentable, Codable, HasClaims, ClaimContainer {
-    public typealias RawValue = [String:Any]
-    public let rawValue: RawValue
+public struct TokenInfo: Codable, HasClaims, ClaimContainer {
```
stevelind-okta, line 20: `HasClaims` seems odd for a protocol name, sound like a bool var

stevelind-okta (approved):
Looks good. Mostly some style comments or suggestions for existing code. I'm missing some context so feel free to ignore any ignorant remarks 👍

README.md
mikenachbaur-okta: Heh yes, I agree. The balance of brevity vs best practices for samples. Here, I erred on the side of code readability

README.md
mikenachbaur-okta: This is still something I'm going back-and-forth on. The goal is to homogenize the interaction patterns for the various authentication flows.  Some flows take a single call, some take one or more calls.  I decided to follow Apple's URLSession pattern where starting, or resuming, a URLSessionDataTask are both invoked using the same method.

Sources/AuthFoundation/AuthFoundation.docc/ManagingUserCredentials.md
mikenachbaur-okta: It throws if there is an error thrown within the underlying storage mechanism.  For example, deserialization fails, Keychain operations fail, etc.

Sources/AuthFoundation/JWT/ClaimContainer.swift
mikenachbaur-okta: It's a container of one or more claims.  There's an enum called `Claim` that represents a single claim.  So perhaps this should be named `ClaimCollection` instead.

Sources/AuthFoundation/AuthFoundation.docc/ManagingUserCredentials.md
mikenachbaur-okta: To answer your other question, these are class functions since at the point a developer would be using this, they wouldn't have an instance of anything to start from.  AuthenticationFlows are only created for the purpose of authenticating, but throughout the lifetime of an app, that only accounts for a brief moment. The majority of the time, the developer will be interacting with credentials.

Plus, they may have multiple credentials for different purposes (some only used infrequently).  So these methods are used to query for / find the credential they need within that part of their application.

Sources/AuthFoundation/JWT/ClaimContainer.swift
mikenachbaur-okta: I have a ticket later on in the process to implement swift lint.  :-D

Sources/AuthFoundation/Token Management/Token+Metadata.swift
```diff
@@ -0,0 +1,94 @@
...
+            if let idToken = token.idToken {
+                let tokenComponents = JWT.tokenComponents(from: idToken.rawValue)
+                if tokenComponents.count == 3 {
+                   payloadData = Data(base64Encoded: tokenComponents[1])
+                }
+            }
+            
+            if let payloadData = payloadData,
+               let payloadInfo = try? JSONSerialization.jsonObject(with: payloadData) as? [String:Any]
+            {
```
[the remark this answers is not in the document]
mikenachbaur-okta: I'm not a fan of the same-line brace when using a multi-line condition. The brace, as well as the code within the brace, get cluttered up with the condition and hurts readability.

Sources/AuthFoundation/User Management/Credential.swift
```diff
@@ -67,9 +67,28 @@ public class Credential {
         try coordinator.with(token: token)
     }
     
+    /// Returns the ``Credential`` that matches the given ID.
+    /// - Parameter id: ID for the credential to return.
+    /// - Returns: Credential matching the ID.
+    public static func with(id: String) throws -> Credential? {
+        try coordinator.with(id: id)
```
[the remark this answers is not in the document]
mikenachbaur-okta: I'm utilizing static dependency injection here for the purposes of testability. There's no occasion where a developer would need to override this particular mechanism, but I also try to avoid using singletons wherever possible. In this case, we can't avoid using static/singleton functions, so I use this dependency injection to be able to more reliably test both sides of the singleton.

PR #35, opened 2022-04-06
Update README documentation

mikenachbaur-okta:
This PR contains updates to README documentation, fixes to unit tests to properly tear down in-memory objects, and update the tvOS sample app to show a QR code.


README.md
```diff
@@ -5,6 +5,15 @@
 
 # Okta Mobile SDK for Swift
 
+The Okta Mobile SDK replaces our legacy mobile SDKs (such as okta-oidc-ios) and enables:
```
stevelind-okta, line 8: Enumerate which legacy mobile SDKs are replaced to avoid confusion? Or maybe qualify the statement in some way so they don't think the DeviceSDK or other swift sdk's are being replaced.

README.md
```diff
@@ -5,6 +5,15 @@
...
+
+* Streamlined development
+* Increased maintainability of the Okta portion of your code
+* Easier feature expansion
+* Implementation of use cases that were previously difficult or impractical
```
stevelind-okta, line 13: Maybe something like "Why you should migrate off the existing SDK"

README.md
```diff
@@ -51,20 +62,23 @@ This SDK is being actively developed and is in Beta release status. At this time
...
-| Login and logout via web redirect | ✅ |
-| Credential management | ✅ |
-| Secure token storage | ✅ |
-| Native SSO / Token Exchange Flow | ✅ |
-| Device Authorization Grant Flow | ✅ |
-| Resource Owner Flow | ✅ |
-| Automatically authorized URLSessionConfiguration for authorizing requests | ✅ |
-| Time synchronization via NTP | 🚧 |
-| Advanced logging and tracing | 🚧 |
-| Transparent upgrades from okta-oidc-ios | 🚧 |
```
stevelind-okta, line 67: Good call not to publish too much here 😄

README.md
```diff
@@ -51,20 +62,23 @@ This SDK is being actively developed and is in Beta release status. At this time
...
+### Key Features
+
+Several key features and capabilities are introduced with this library, with some notable improvements listed below.
```
stevelind-okta, line 67: Improvements over what? Hopefully new users will just use this instead of looking at the old SDK, yes?

Tests/AuthFoundationTests/KeychainTokenStorageTests.swift
```diff
@@ -17,7 +17,7 @@ import XCTest
 @testable import TestCommon
 
 final class KeychainTokenStorageTests: XCTestCase {
-    let mock = MockKeychain()
+    var mock: MockKeychain!
```
stevelind-okta, line 20: No objection to this change, in the past we have done this to avoid long-lived variables as the test target may keep these values in memory for the life entire test suite (as opposed to class by class)

stevelind-okta (approved):

README.md
mikenachbaur-okta: Good observation. I originally had a list of SDKs this is replacing, but didn't want to "spook the wildlife" since we don't have release estimates or notices of deprecation yet.  We'll come up with a definitive list a bit later, once we identify a migration plan.

README.md
mikenachbaur-okta: Yes, this section will be refined / replaced later, once we get closer to a 1.0 release. This README, for now, is largely targeting our prospective beta users.  Once this is released fully, we'll update the other SDK's READMEs to point here, and describe why they might want to use our new SDK.

Tests/AuthFoundationTests/KeychainTokenStorageTests.swift
mikenachbaur-okta: Exactly the reason these test changes are being made.  XCTestCase instances last for the whole lifetime of the test run, so I'm just making it a common practice to make any test properties that maintain state to be made var, and setup and torn down for each invocation.

PR #36, opened 2022-04-07
Add initial swiftlint file, and fix the current violations

mikenachbaur-okta:


stevelind-okta (approved):
🚢 Thanks, really appreciate this change

PR #38, opened 2022-04-20
Fixes for IDX integration

alexnachbaur-okta:


Sources/AuthFoundation/User Management/Credential.swift
```diff
@@ -188,11 +188,20 @@ public class Credential: Identifiable, Equatable, OAuth2ClientDelegate {
     ///   - type: The token type to revoke.
     ///   - completion: Completion block called when the operation completes.
     public func revoke(type: Token.RevokeType = .accessToken, completion: ((Result<Void, OAuth2Error>) -> Void)? = nil) {
+        var shouldRemove = false
```
jaynewstrom-okta, line 191: Should `revoke` remove the token? The kotlin SDK doesn't do this.

mikenachbaur-okta: This is related to [OKTA-467618](https://oktainc.atlassian.net/browse/OKTA-467618). If the token is no longer valid / cannot be refreshed, it's automatically removed to prevent inadvertent errors.

This change is updating the logic to take account for the fact that an access token can be recreated if a valid refresh token is present.  However, if you revoke the refresh token, both the refresh and access tokens are invalidated. https://developer.okta.com/docs/guides/revoke-tokens/main/#revoke-only-the-refresh-token

PR #39, opened 2022-04-21
Control biometric / keychain access control settings

mikenachbaur-okta:


Sources/AuthFoundation/Security/Internal/Keychain+Extensions.swift
```diff
@@ -14,13 +14,158 @@
...
+                return false
+            }
+        }
+    }
+
+    func performUpdate(_ item: Keychain.Item, authenticationContext: KeychainAuthenticationContext?) throws {
+        let updateSearchQuery = self.updateQuery
+
+        var saveQuery = item.query
+        saveQuery.removeValue(forKey: kSecClass as String)
```
IldarAbdullin-okta, line 123: BTW, why we are doing remove if our underlying call is `SecItemUpdate`? User may loose previous data if `SecItemUpdate` fails

Sources/AuthFoundation/Security/Keychain.swift
```diff
@@ -48,27 +61,50 @@ public struct Keychain {
...
+                cfDictionary[kSecUseAuthenticationContext as String] = authenticationContext
+            }
+            
+            if let accessControl = accessControl {
+                cfDictionary[kSecAttrAccessControl as String] = accessControl
+                cfDictionary.removeValue(forKey: kSecAttrAccessible as String)
+            }
 
-            Keychain.implementation.deleteItem(cfDictionary)
+            Keychain.implementation.deleteItem(cfDictionary as CFDictionary)
```
IldarAbdullin-okta, line 82: Maybe just create explicit error for duplicate item(errSecDuplicateItem) case instead of physically delete? There is a risk that use may loose the data

Sources/AuthFoundation/Token Management/Internal/KeychainTokenStorage.swift
```diff
@@ -87,69 +96,74 @@ class KeychainTokenStorage: TokenStorage {
...
+                                    accessControl: accessControl,
+                                    synchronizable: accessibility.isSynchronizable,
+                                    label: nil,
+                                    description: nil,
+                                    generic: nil,
+                                    value: data)
         
-        delegate?.token(storage: self, replaced: id, from: oldToken, to: token)
+        try? oldResult.delete()
+        try newItem.save(authenticationContext: security?.context)
```
IldarAbdullin-okta, line 127: Why we don't use `update` API here?

Sources/AuthFoundation/Token Management/Internal/KeychainTokenStorage.swift
```diff
@@ -87,69 +96,74 @@ class KeychainTokenStorage: TokenStorage {
...
-        
-        try Keychain
+        let item = Keychain
             .Item(account: metadata.id,
                   service: KeychainTokenStorage.metadataName,
                   accessibility: .afterFirstUnlock,
                   value: try encoder.encode(metadata))
-            .save()
+
+        try? item.delete()
```
IldarAbdullin-okta, line 165: Risky. Maybe just build the logic around Add/Update APIs rather than Delete/Add? I think `Add/Update` is not that risky, at least can't cause data loss

IldarAbdullin-okta (approved):
Very clean and readable 👍 
Left several comments to consider

Sources/AuthFoundation/Security/Internal/Keychain+Extensions.swift
```diff
@@ -14,13 +14,158 @@
...
+            case kSecAttrAccessGroup: fallthrough
+            case kSecAttrAccount:
+                return true
+            default:
+                return false
+            }
+        }
+    }
+
+    func performUpdate(_ item: Keychain.Item, authenticationContext: KeychainAuthenticationContext?) throws {
```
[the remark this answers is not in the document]
mikenachbaur-okta: This is calling the underlying keychain update API.

Sources/AuthFoundation/Token Management/Internal/KeychainTokenStorage.swift
mikenachbaur-okta: Update has some limitations, in particular it doesn't let you return a reference of the newly created keychain item. So for cases such as this, a delete/remove is often safer. 

I plan to use `update` in the setMetadata method however.

Sources/AuthFoundation/Token Management/Internal/KeychainTokenStorage.swift
mikenachbaur-okta: I mentioned this a bit up above, both the discussion around how to handle Delete/Add, as well as the use-cases around Update. For Metadata items, I'll change it to use Update, since I don't need to get the resulting item.

Sources/AuthFoundation/Security/Internal/Keychain+Extensions.swift
```diff
@@ -14,13 +14,158 @@
...
+        let status = Keychain.implementation.copyItemMatching(cfQuery as CFDictionary, &ref)
+        guard status == noErr else {
+            if status == errSecItemNotFound {
+                throw KeychainError.notFound
+            } else {
+                throw KeychainError.cannotGet(code: status)
+            }
+        }
+        
+        guard let result = ref as? Dictionary<String, Any> else {
```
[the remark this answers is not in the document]
IldarAbdullin-okta: Ok, I think I missed important setting in query `kSecReturnAttributes` so you basically asking keychain to return attributes along with data. So downcast to `Dictionary<String, Any>` a bit confused me as without `kSecReturnAttributes` setting keychain returns reference to Data or SecKey type

Sources/AuthFoundation/Token Management/Internal/KeychainTokenStorage.swift
```diff
@@ -87,69 +96,74 @@ class KeychainTokenStorage: TokenStorage {
...
         token.id = id
         
         let data = try encoder.encode(token)
-        try Keychain
-            .Item(account: id,
-                  service: KeychainTokenStorage.serviceName,
-                  accessibility: .unlocked,
-                  value: data)
-            .save()
+        let accessibility = security?.accessibility ?? .afterFirstUnlock
```
IldarAbdullin-okta, line 112: Should we derive this setting from `oldResult`?

Sources/AuthFoundation/Token Management/Internal/KeychainTokenStorage.swift
```diff
@@ -87,69 +96,74 @@ class KeychainTokenStorage: TokenStorage {
...
+        let accessGroup = security?.accessGroup
```
IldarAbdullin-okta, line 113: Should we derive this setting from `oldResult`?

Sources/AuthFoundation/Token Management/Internal/KeychainTokenStorage.swift
IldarAbdullin-okta: You basically already did the GET of oldResult so you have all required attributes to copy into the new Item. I still like the GET/UPDATE approach since it is only 2 operations comparing to GET/DELETE/ADD. Data safety aspect is also important

mikenachbaur-okta: You're right, in this particular use-case you're totally correct, this is a prime candidate for Update.  I'll make that change.

Sources/AuthFoundation/Token Management/Internal/KeychainTokenStorage.swift
mikenachbaur-okta: The user could potentially change the accessibility settings, but you're correct in that it should default to the current value.

PR #40, opened 2022-04-27
Ensure metadata always accompanies tokens in the keychain, and use 'update' to change metadata

mikenachbaur-okta:


Samples/WebSignIn/Common (iOS)/Keychain/Keychain+Token.swift
```diff
@@ -22,17 +22,23 @@ public extension Keychain {
...
-                .map{ try $0.save() }
+        try [Token.Kind.idToken, Token.Kind.deviceSecret]
+            .forEach { try saveDeviceSSO(token, kind: $0) }
+    }
+    
+    static func saveDeviceSSO(_ token: Token, kind: Token.Kind) throws {
+        guard let tokenString = token.token(of: kind),
+              let data = tokenString.data(using: .utf8)
+        else {
+            return
```
IldarAbdullin-okta, line 33: Throw?

IldarAbdullin-okta (approved):
LGTM!

Samples/WebSignIn/Common (iOS)/Keychain/Keychain+Token.swift
```diff
@@ -22,17 +22,23 @@ public extension Keychain {
...
         }
+        
+        try Keychain
+            .Item(account: "Okta-\(kind.rawValue)",
+                  service: "Okta",
```
jmelberg-okta, line 38: Do these need to be namespaced to Okta? May be worthwhile to make this configurable by the consumer, as it'll delegate global keychain management/migration to the consumer.

Samples/WebSignIn/Common (iOS)/Keychain/Keychain+Token.swift
mikenachbaur-okta: This is a utility function for the sample app, so it is t a high priority.

Samples/WebSignIn/Common (iOS)/Keychain/Keychain+Token.swift
```diff
@@ -22,17 +22,23 @@ public extension Keychain {
...
-                                  service: "Okta",
-                                  accessGroup: oktaMobileSdkAccessGroup,
-                                  value: $0)
-                }
-                .map{ try $0.save() }
+        try [Token.Kind.idToken, Token.Kind.deviceSecret]
+            .forEach { try saveDeviceSSO(token, kind: $0) }
+    }
+    
+    static func saveDeviceSSO(_ token: Token, kind: Token.Kind) throws {
```
jmelberg-okta, line 29: Curious - can we save multiple tokens of the same `Kind`? Am assuming that our `account` value below is our distinguishing characteristic

Samples/WebSignIn/Common (iOS)/Keychain/Keychain+Token.swift
mikenachbaur-okta: This is just for the purposes of the sample app. A real app wouldn't do things in this fashion.

Samples/WebSignIn/Common (iOS)/Keychain/Keychain+Token.swift
mikenachbaur-okta: You'll want to check out this other PR that extended my Keychain implementation for more information https://github.com/okta/okta-mobile-swift/pull/39

Ultimately this class is very flexible and clean, so it should be useful for you. 

This use-case in the sample app is only there to save the device secret and ID token with an accompanying sample app to demonstrate Native SSO.

PR #41, opened 2022-05-03
Add support for migrating from OktaOidc

mikenachbaur-okta:
* Adds the concept of an SDKVersionMigrator, to attempt to future-proof version migrations that may need to occur later.
* Centralizes Plist configuration loading for better reuse.
* Sample application to demonstrate migration from OktaOidc.
* Fixes a bug in how token `issuedAt` timestamps are created.

Tests to come in a subsequent commit to this PR.


Sources/AuthFoundation/Migration/Migration.swift
```diff
@@ -0,0 +1,82 @@
...
+    /// Namespace used for a variety of version migration agents.
+    public class Migration {
+        fileprivate static var hasMigrated: Bool = false
+        static var shared: Migration = {
+            Migration()
+        }()
+        
+        fileprivate(set) static var registeredMigrators: [SDKVersionMigrator] = []
+        
+        struct State: Codable {
```
mikenachbaur-okta, line 55: Unused currently, will rip it out in my next PR.

Samples/OIDCMigration/OIDCMigration/OIDCSignInViewController.swift
```diff
@@ -0,0 +1,136 @@
...
+            migrateButton.isHidden = false
+            openUserProfileButton.isHidden = true
+        } else {
+            migrateButton.isHidden = true
+            openUserProfileButton.isHidden = false
+        }
+    }
+
+    @IBAction func ephemeralSwitchChanged(_ sender: Any) {
+        guard let sender = sender as? UISwitch else { return }
```
stevelind-okta, line 72: does it matter which switch is the sender?

Sources/AuthFoundation/Migration/Migration.swift
```diff
@@ -0,0 +1,82 @@
...
+    /// Version migrators are utilized to migrate user data on an as-needed basis.
+    /// - Parameter migrator: Migrator to register.
+    public static func register(migrator: SDKVersionMigrator) {
+        Migration.registeredMigrators.append(migrator)
+    }
+    
+    /// Namespace used for a variety of version migration agents.
+    public class Migration {
+        fileprivate static var hasMigrated: Bool = false
+        static var shared: Migration = {
```
mikenachbaur-okta, line 49: I want to deallocate this shared instance after migration has completed, but it introduces some interesting race conditions.  I might change this later, but for now I don't want to prematurely optimize things.

Samples/OIDCMigration/OIDCMigrationUITests/OIDCMigrationUITestsLaunchTests.swift
```diff
@@ -0,0 +1,37 @@
...
+        true
+    }
+
+    override func setUpWithError() throws {
+        continueAfterFailure = false
+    }
+
+    func testLaunch() throws {
+        let app = XCUIApplication()
+        app.launch()
```
stevelind-okta, line 27: Are you planning to add UITests for this change?

Sources/AuthFoundation/Migration/Migrators/OIDCLegacyMigrator.swift
```diff
@@ -0,0 +1,270 @@
...
+                              accessToken: accessToken,
+                              scope: scope,
+                              refreshToken: tokenResponse.refreshToken,
+                              idToken: idToken,
+                              deviceSecret: nil,
+                              context: Token.Context(configuration: configuration,
+                                                     clientSettings: clientSettings))
+            
+            var security = Credential.Security.standard
+            if let accessibility = item.accessibility {
```
mikenachbaur-okta, line 188: Preserving the keychain accessibility / access control settings from the previous keychain.

Sources/AuthFoundation/Migration/Migration.swift
```diff
@@ -0,0 +1,82 @@
...
+//
+// See the License for the specific language governing permissions and limitations under the License.
+//
+
+import Foundation
+
+public enum SDKVersionMigrationError: Error {
+    case cannotCreateMigrator
+    case configurationNotFound
+    case configurationInvalid
```
stevelind-okta, line 18: add description docs for these?

Sources/AuthFoundation/Network/Internal/String+AuthFoundation.swift
```diff
@@ -54,7 +54,7 @@ private let systemVersion: String = {
     #endif
 }()
 
-public class SDKVersion {
+public class SDKVersion: Codable {
```
mikenachbaur-okta, line 57: Will revert this; Codable was to be used when tracking which versions had been migrated, but that seems to be unnecessary at this point.

Sources/AuthFoundation/Migration/Migrators/OIDCLegacyMigrator.swift
```diff
@@ -0,0 +1,270 @@
...
+                    let item = try searchResult.get()
+                    return (searchResult, try decode(item.value))
+                }).forEach({ (item: Keychain.Search.Result, oldModel: StateManager?) in
+                    try importToken(item, from: oldModel)
+                })
+        }
+
+        private func decode(_ data: Data) throws -> StateManager? {
+            let archiver: NSKeyedUnarchiver
+            if #available(iOS 11.0, *) {
```
stevelind-okta, line 126: maybe consider dropping support below 11?

Sources/AuthFoundation/OAuth2/OAuth2ClientConfiguration.swift
```diff
@@ -0,0 +1,113 @@
...
+        public let clientId: String
+        public let scopes: String
+        public let redirectUri: URL?
+        public let logoutRedirectUri: URL?
+        public let additionalParameters: [String: String]?
+
+        public init() throws {
+            guard let file = Bundle.main.url(forResource: "Okta", withExtension: "plist") else {
+                throw PropertyListConfigurationError.defaultPropertyListNotFound
+            }
```
stevelind-okta, line 37: extension on `Bundle`?

Sources/AuthFoundation/OAuth2/OAuth2ClientConfiguration.swift
```diff
@@ -0,0 +1,113 @@
...
+            
+            let redirectUri: URL?
+            if let redirectUriString = dict["redirectUri"] {
+                redirectUri = URL(string: redirectUriString)
+            } else {
+                redirectUri = nil
+            }
+            
+            let logoutRedirectUri: URL?
+            if let logoutRedirectUriString = dict["logoutRedirectUri"] {
```
stevelind-okta, line 79: could use an enum for these keys, but I'm not against short strings (up to you)

stevelind-okta (commented):
Looks good. Left some minor comments.

In the devices SDK we are trying to move to SQLite instead of key-value stores, as it's harder to reason about migrations from a KV (as you probably found). Do you think it is something to consider? Unfortunately adding a DB would likely add dependencies, so it's not a light option.

Samples/OIDCMigration/OIDCMigration/OIDCSignInViewController.swift
mikenachbaur-okta: No, not at this time. It can be refactored later as needed.

Samples/OIDCMigration/OIDCMigrationUITests/OIDCMigrationUITestsLaunchTests.swift
mikenachbaur-okta: Yes, every one of the sample apps will have E2E tests for them.

Sources/AuthFoundation/Migration/Migrators/OIDCLegacyMigrator.swift
mikenachbaur-okta: The SDK currently supports versions down to iOS 9, so backwards compatibility is still important.

Sources/AuthFoundation/OAuth2/OAuth2ClientConfiguration.swift
mikenachbaur-okta: A subsequent PR later on will change the names of these keys to be more Apple-esque, but the older (e.g. these current) keys will still be needed for backward compatibility.  So having enum-based keys might make that challenging.

Sources/AuthFoundation/Migration/Migrators/OIDCLegacyMigrator.swift
stevelind-okta: Wow. Does the new SDK still need to keep that level of support?

Samples/OIDCMigration/OIDCMigration/OIDCSignInViewController.swift
```diff
@@ -0,0 +1,136 @@
...
+                signInButton.isHidden = true
+                signOutButton.isHidden = false
+            } else {
+                signInButton.isHidden = false
+                signOutButton.isHidden = true
+            }
+        } else {
+            clientIdLabel.text = "Not configured"
+            signInButton.isEnabled = false
+            signOutButton.isHidden = true
```
zeroes-ones, line 58: can we extract these into function, in this way we can reuse

Samples/OIDCMigration/OIDCMigration/SceneDelegate.swift
```diff
@@ -0,0 +1,57 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//
+// See the License for the specific language governing permissions and limitations under the License.
+//
+
+import UIKit
+
+class SceneDelegate: UIResponder, UIWindowSceneDelegate {
```
zeroes-ones, line 15: I think we can safely delete scene delegate, if not using it?

mikenachbaur-okta:
@stevelind-okta I agree that SQLite is much more flexible for storing configuration and application state, but doing so would limit our ability to support iCloud Keychain Sharing, persistent configuration after application deletion, etc. Going forward my approach is to be more intentional about the values we store in the keychain, to make it easier to uniquely identify credentials used for migration.

Samples/OIDCMigration/OIDCMigration/OIDCSignInViewController.swift
mikenachbaur-okta: There doesn't seem to be much of a need for reuse within the sample apps. My intention was to keep the flow clear to understand, and to avoid adding additional complexity that could distract from the sample itself.

Sources/AuthFoundation/Migration/Migration.swift
mikenachbaur-okta: I have stories for next sprint to expand doc coverage.

Samples/OIDCMigration/OIDCMigration/SceneDelegate.swift
mikenachbaur-okta: I generally agree, but in the case of the sample apps, I like to keep them as close to what Apple provides in their templates. I wouldn't want to imply to developers that the SDK _doesn't_ work with a SceneDelegate, for example.

Sources/AuthFoundation/OAuth2/OAuth2ClientConfiguration.swift
mikenachbaur-okta: Since this is associated with the struct itself, it made sense to me to make this an initializer on the configuration object itself, instead of separating it out to a different extension.

Sources/AuthFoundation/Migration/Migrators/OIDCLegacyMigrator.swift
mikenachbaur-okta: FYI @stevelind-okta you might be interested to see our [OS version support policy](https://github.com/okta/okta-mobile-swift#platforms), the key feature of which is that we don't remove support for an OS version unless there's a compelling reason not to do so.  We only support down to iOS 11 officially, and dropping support for older versions isn't considered a breaking change.

Sources/WebAuthenticationUI/Resources/en.lproj/Localizable.strings
```diff
@@ -1,8 +1,4 @@
 /* WebAuthenticationError */
-"default_property_list_not_found_description" = "The default Okta.plist configuration file was not found.";
-"invalid_property_list_description" = "The configuration file named \"%@\" was invalid.";
-"cannot_parse_property_list_description" = "Could not parse the configuration file named \"%@\".";
-"missing_configuration_values_description" = "Web authentication is not configured properly.";
```
IldarAbdullin-okta, line 5: Not completely understand the idea of using localizable strings in SDK. Are you going to translate all these strings? Or you want to delegate this to app developer?
How you plan to deploy localization strings using Cocoapods? Pack into resource bundle?

IldarAbdullin-okta (approved):
Overall looks good. Left several comments

Samples/OIDCMigration/OIDCMigration/OIDCSignInViewController.swift
```diff
@@ -0,0 +1,136 @@
...
+    
+    var oktaOidc: OktaOidc?
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        
+        updateUIState()
+        
+        // Dismiss the profile view controller when it is signed out
+        NotificationCenter.default.addObserver(forName: .defaultCredentialChanged, object: nil, queue: .main) { note in
```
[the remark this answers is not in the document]
mikenachbaur-okta: Given this is a sample app, I'm trying to keep things simple. But you're right, we should follow best practices.

IldarAbdullin-okta:
> @stevelind-okta I agree that SQLite is much more flexible for storing configuration and application state, but doing so would limit our ability to support iCloud Keychain Sharing, persistent configuration after application deletion, etc. Going forward my approach is to be more intentional about the values we store in the keychain, to make it easier to uniquely identify credentials used for migration.

IMO, add sqlite into these SDKs is overkill. We added sqlite into Device SDK primarily due to Okta Verify non-trivial use-cases - support of many enrollments(multi-org, multi-user). For CIAM case it will be mainly one enrollment

mikenachbaur-okta:
Furthermore @IldarAbdullin-okta, introducing SQLite brings in compliance issues, crypto of data-at-rest issues, and a whole host of other problems.  Plus having to manage the lifecycle of opening and closing connections to the database.

If we decide to expand upon our Linux compatibility, I could see SQLite as an alternative to keychain-based storage though.

Sources/WebAuthenticationUI/Resources/en.lproj/Localizable.strings
mikenachbaur-okta: One of the issues we've faced with client SDKs in the past is developers blindly printing errors to the user, which results in very sub-optimal developer experiences.  Either that, or the error messages themselves are confusing to developers as they debug the application.

However, I'm not a fan of hard-coding strings when it's avoidable, so I'm starting with localization of these errors.  We may decide to roll this back in the future.

As far as Cocoapods support is concerned, yes my plan is to roll these into a resource bundle.

zeroes-ones (approved):

stevelind-okta (approved):
Nice tests 👍

PR #45, opened 2022-05-11
Add Device Authorization Flow E2E UI tests

mikenachbaur-okta:
This also updates / clarifies the minimum supported OS versions for the various supported platforms, as well as formatted it into a clearer to read table.


README.md
```diff
@@ -236,13 +236,15 @@ The minimum supported Swift 5 minor version is the one released with the oldest-
...
 ### Platforms
 
-Only the last 4 major platform versions are officially supported, starting from:
-
-- iOS 12
-- macOS 10.15
-- Catalyst 13
-- tvOS 12
-- watchOS 6.2
+Only the last 4 major platform versions are officially supported, unless there are platform limitations that limits our ability to support older versions.
```
jaynewstrom-okta, line 239:
```suggestion
Only the last 4 major platform versions are officially supported, unless there are platform limitations that limit our ability to support older versions.
```

This still doesn't read easily. Maybe use prevent instead of limit?

jaynewstrom-okta (approved):

PR #47, opened 2022-05-12
Update CI jobs for unit and UI tests

mikenachbaur-okta:


.github/workflows/ui-tests.yaml
```diff
@@ -0,0 +1,101 @@
...
+      - dev-*
+      - release-*
+  pull_request:
+    branches:
+      - master
+      - dev-*
+      - release-*
+
+env:
+  DEVELOPER_DIR: /Applications/Xcode_13.3.app/Contents/Developer
```
moisesolmedo-okta, line 16: will this break for future releases? also couldn't find where it's being used. Is it needed?

moisesolmedo-okta (approved):

.github/workflows/ui-tests.yaml
mikenachbaur-okta: When you set that environment variable, `xcodebuild` will automatically stitch to the version at that path temporarily. 

And in the future I plan to run against `latest` once GitHub Actions switches it's default.

PR #48, opened 2022-05-16
Removed unused errors, and added more localizable descriptions

mikenachbaur-okta:


Sources/AuthFoundation/Token Management/TokenError.swift
```diff
@@ -0,0 +1,48 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//
+// See the License for the specific language governing permissions and limitations under the License.
+//
+
+import Foundation
+
+public enum TokenError: Error {
```
JayNewstrom, line 15: Should we add docs here?

JayNewstrom (approved):

jaynewstrom-okta (approved):

Sources/AuthFoundation/Token Management/TokenError.swift
mikenachbaur-okta: I haven't been adding header docs for the error messages, at least not yet.

PR #63, opened 2022-06-15
Fix a bug where signing-out and then -in will open two sign in browsers

mikenachbaur-okta:
This should fix your issue @moisesolmedo-okta


jaynewstrom-okta (approved):
Can you add a test for the fixed behavior?

mikenachbaur-okta:
> Can you add a test for the fixed behavior?

Yes, that's on the docket for this morning. I wanted to get @moisesolmedo-okta's eyes on this early on to see if it addressed his problems.

moisesolmedo-okta (approved):
👍

PR #64, opened 2022-06-15
Simplify the Credential.refresh* methods

mikenachbaur-okta:


mauricesharp-okta:
Looks good to me.

Sources/AuthFoundation/OAuth2/OAuth2Client.swift
```diff
@@ -246,6 +246,10 @@ public class OAuth2Client {
                         case .failure(let error):
                             action.finish(.failure(.network(error: error)))
                             self.delegateCollection.invoke { $0.oauth(client: self, didRefresh: token, replacedWith: nil) }
+                            
+                            NotificationCenter.default.post(name: .tokenRefreshFailed,
+                                                            object: token,
+                                                            userInfo: ["error": error])
```
stevelind-okta, line 252: would it make sense to move this note to the delegateCollection?

Sources/AuthFoundation/User Management/Credential.swift
```diff
@@ -181,27 +185,29 @@ public class Credential: Identifiable, Equatable, OAuth2ClientDelegate {
...
-                                completion: ((Result<Token, OAuth2Error>) -> Void)? = nil)
+                                completion: ((Result<Void, OAuth2Error>) -> Void)? = nil)
     {
         if let expiresAt = token.expiresAt,
             expiresAt.timeIntervalSinceNow <= graceInterval
         {
             refresh(completion: completion)
         } else {
-            completion?(.success(token))
+            completion?(.success(()))
```
stevelind-okta, line 210: looks funny 😄

stevelind-okta (approved):
I guess the main idea is that devs don't usually need to get the token, so refresh is mostly used at `let _` type call?

Sources/AuthFoundation/OAuth2/OAuth2Client.swift
mikenachbaur-okta: The delegateCollection is used in a number of different areas, and I wouldn't want to add unnecessary overhead elsewhere.

Sources/AuthFoundation/User Management/Credential.swift
mikenachbaur-okta: Heh yeah, the Swift syntax for passing void to a Result is definitely odd. I wish you could just leave it blank...

mikenachbaur-okta:
Correct. If a developer really wants to deal with plain access tokens, they can use the Token object directly and don't need the Credential convenience class. If they are using the conveniences though, the recommended practice is to just use the `authorize(_:)` method to add authorization headers to a URLRequest.

The doc feedback I got was that it was confusing, during a refresh, what to do with the token. So it's clearer to just not return it, since there's nothing actionable on the part of a developer.

PR #67, opened 2022-06-22
Add article on authenticator flows

mikenachbaur-okta:


mikenachbaur-okta:
> Quick question: did you run the diagrams past Alex Cohen?

I didn't, I first wanted to see if this sort of article makes sense in the DocC content. 

I'll check those figma templates out, and will pass the resulting diagrams past Alex. Thanks for the tip!

mauricesharp-okta:
This is a great set of information for developers using the SDK. The first thing that occurred to me when I read it is that it should be in the developer.okta.com documentation: Guides > Sign users in > Mobile authentication > Authentication using OAuth2 (or some appropriate title).

If you're OK with that, I'd like to use this content as the basis for a new article.

mikenachbaur-okta:
After a conversation with @mauricesharp-okta it looks like this is ready to merge.

emmanuelogunmuyiwa-okta (approved):

PR #74, opened 2022-07-05
Update device authorization and JWKS to closer match the spec

mikenachbaur-okta:
This matches up to https://github.com/okta/okta-mobile-kotlin/pull/183 and https://github.com/okta/okta-mobile-kotlin/pull/182

### DIFF
--- a/Tests/AuthFoundationTests/TokenTests.swift
+++ b/Tests/AuthFoundationTests/TokenTests.swift
@@ -12,6 +12,7 @@
 
 import XCTest
 @testable import AuthFoundation
+@testable import TestCommon
 
 final class TokenTests: XCTestCase {
     let configuration = OAuth2Client.Configuration(baseURL: URL(string: "https://example.com")!,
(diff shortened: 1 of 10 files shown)


JayNewstrom (commented):
I think the one missing thing from those 2 PRs is default scope.

https://github.com/okta/okta-mobile-swift/blob/master/Sources/AuthFoundation/Token%20Management/Token.swift#L168

I choose null, but there might be a more correct answer.

The rest looks good though.

mikenachbaur-okta:
Yeah, that seems like the right thing to do. Also, Okta has a feature for defining a default scope for when a client doesn't supply one, so we might want to make the `scope` argument optional for our authorization flows.

Tests/AuthFoundationTests/TokenTests.swift
```diff
@@ -47,6 +48,22 @@ final class TokenTests: XCTestCase {
         XCTAssertEqual(context, decodedContext)
     }
     
+    func testNilScope() throws {
+        let data = data(for: """
+        {
+           "token_type": "Bearer",
+           "expires_in": 3600,
```
JayNewstrom, line 55: Worth noting, some Authorization servers may not return expires_in (notably when you are doing pure OAuth, without OIDC). I don't think it's something we need to account for, but I think the only 2 things that are required from OAuth are access_token/token_type.

mikenachbaur-okta: Yeah, I'm not sure if we should worry about that. There's a lot of code that assumes the `expires_in` exists, for both Swift and Kotlin SDKs.  So I'd hold off on this, until it's necessary.

PR #75, opened 2022-07-06
Support the `slow_down` token error response in DeviceAuthorizationFlow

mikenachbaur-okta:
Corresponds to https://github.com/okta/okta-mobile-kotlin/pull/186

### DIFF
--- a/Sources/OktaOAuth2/Authentication/DeviceAuthorizationFlow.swift
+++ b/Sources/OktaOAuth2/Authentication/DeviceAuthorizationFlow.swift
@@ -320,11 +320,6 @@ extension DeviceAuthorizationFlow: UsesDelegateCollection {
 }
 
 extension DeviceAuthorizationFlow {
-    struct TimerInfo {
-        let context: Context
-        let completion: ((Result<Token, APIClientError>) -> Void)?
-    }
-    
     func getToken(using context: Context, completion: @escaping(Result<Token?, APIClientError>) -> Void) {
         client.openIdConfiguration { result in
             switch result {
@@ -366,10 +361,6 @@ extension DeviceAuthorizationFlow {
                 self.delegateCollection.invoke { $0.authentication(flow: self, received: error) }
             }
         }
-    }
-    
-    func finish(_ result: Result<Token?, APIClientError>) {
-        
     }
 }
 
--- a/Sources/OktaOAuth2/Authentication/DeviceAuthorizationFlow.swift
+++ b/Sources/OktaOAuth2/Authentication/DeviceAuthorizationFlow.swift
@@ -67,7 +67,7 @@ public final class DeviceAuthorizationFlow: AuthenticationFlow {
     /// A model representing the context and current state for an authorization session.
     public struct Context: Decodable, Equatable, Expires {
         let deviceCode: String
-        let interval: TimeInterval
+        var interval: TimeInterval
         
         /// The date this context was created.
         public let issuedAt: Date?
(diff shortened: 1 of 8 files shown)


Tests/OktaOAuth2Tests/DeviceAuthorizationFlowErrorTests.swift
```diff
@@ -0,0 +1,134 @@
...
+        Token.accessTokenValidator = MockTokenHashValidator()
+        flow = client.deviceAuthorizationFlow()
+    }
+    
+    override func tearDownWithError() throws {
+        JWK.resetToDefault()
+        Token.resetToDefault()
+    }
+    
+    func testSlowDown() throws {
```
JayNewstrom, line 41: The down side I see to these tests are we aren't asserting anything about what delay is happening.

Here's the related test in okta-mobile-kotlin, that asserts the timeouts given https://github.com/okta/okta-mobile-kotlin/pull/186/files#diff-ee0407074640dcfed6eb5fa913fb504cbee134b4bdf152f89b9e11efe3175091R285

JayNewstrom (approved):
Overall this looks good. Just one note.

Sources/OktaOAuth2/Authentication/DeviceAuthorizationFlow.swift
```diff
@@ -226,7 +226,37 @@ public final class DeviceAuthorizationFlow: AuthenticationFlow {
...
+        completion = nil
+        isAuthenticating = false
+    }
+
+    // MARK: Private properties / methods
+    static var slowDownInterval: TimeInterval = 5
+    
+    var timer: DispatchSourceTimer?
+    var completion: ((Result<Token, APIClientError>) -> Void)?
+    public let delegateCollection = DelegateCollection<DeviceAuthorizationFlowDelegate>()
```
emmanuelogunmuyiwa-okta, line 247: Do you think It might be worth adding a documentation for this since it's a public api?

Sources/OktaOAuth2/Authentication/DeviceAuthorizationFlow.swift
```diff
@@ -319,14 +336,26 @@ extension DeviceAuthorizationFlow {
...
-                            completion(.failure(error))
+                            switch oauthError.code {
+                            case "authorization_pending":
+                                // Keep polling, since we haven't received a token yet.
+                                completion(.success(nil))
+                                return
+                                
+                            case "slow_down":
+                                // Increase the polling interval for all subsequent requests, according to the specification.
+                                self.scheduleTimer(offsetBy: DeviceAuthorizationFlow.slowDownInterval)
```
emmanuelogunmuyiwa-okta, line 349: It seems like we need to capture  weak reference to `self` in the closure above. any thoughts ?

mikenachbaur-okta: The retain cycle is broken when the flow completes and `reset()` is called, but I see your point.

Tests/OktaOAuth2Tests/DeviceAuthorizationFlowErrorTests.swift
mikenachbaur-okta: Yes, I saw how you did that in your tests. I don't have an easy way of tracking that in my tests but ... wait, never mind, I just thought of a way that might work. 😄  Thanks for letting me rubber-duck on this.

Sources/OktaOAuth2/Authentication/DeviceAuthorizationFlow.swift
emmanuelogunmuyiwa-okta: I see your point and I agree with documenting `UsesDelegateCollection`.

PR #77, opened 2022-07-07
Introduced OAuth2ServerErrorCode to improve OAuth2ServerError and its documentation 

emmanuelogunmuyiwa-okta:
- This provides some context around the known OAuth2 ServerErrorCodes
list of known error codes: https://developer.okta.com/docs/reference/api/oidc/#possible-errors

### DIFF
--- a/Sources/AuthFoundation/Responses/OAuth2ServerError.swift
+++ b/Sources/AuthFoundation/Responses/OAuth2ServerError.swift
@@ -15,7 +15,7 @@ import Foundation
 /// Describes errors reported from an OAuth2 server.
 public struct OAuth2ServerError: Decodable, Error, LocalizedError {
     /// Error code.
-    public let code: String
+    public let code: OAuth2ServerErrorCode
     
     /// Error message, or description.
     public let description: String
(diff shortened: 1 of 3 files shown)


Sources/AuthFoundation/Responses/OAuth2ServerError.swift
```diff
@@ -33,3 +33,35 @@ public struct OAuth2ServerError: Decodable, Error, LocalizedError {
         case description = "errorDescription"
     }
 }
+
+extension OAuth2ServerError {
+    ///  Possible  OAuth 2.0 server error code
+    public enum OAuth2ServerErrorCode: String, Decodable {
```
mikenachbaur-okta, line 39: This name feels a bit verbose. I'd suggest finding something shorter, like "Code".  The rest is implied, since it's already within the `OAuth2ServerError` namespace, so repeating it in the enum name seems redundant.

Sources/AuthFoundation/Responses/OAuth2ServerError.swift
```diff
@@ -33,3 +33,35 @@ public struct OAuth2ServerError: Decodable, Error, LocalizedError {
...
+        /// The authorization request is still pending as the end user hasn't yet completed the user-interaction step
+        case authorization_pending
```
mikenachbaur-okta, line 41: Can you camelCase these enums, to be more "swifty"?

mikenachbaur-okta (changes requested):

Sources/AuthFoundation/Responses/OAuth2ServerError.swift
emmanuelogunmuyiwa-okta: Good callout on this. I agree with you, I'll make the changes.

Sources/AuthFoundation/Responses/OAuth2ServerError.swift
emmanuelogunmuyiwa-okta: Definitely. I've made the change.

mikenachbaur-okta (approved):
Looks good!

PR #80, opened 2022-07-11
Improve readability of DefaultIDTokenValidator flow

emmanuelogunmuyiwa-okta:
-  The goal of this is to Improve readability of DefaultIDTokenValidator and provide a clear understanding of each validation checks performed.

### DIFF
--- a/Sources/AuthFoundation/JWT/Internal/DefaultIDTokenValidator.swift
+++ b/Sources/AuthFoundation/JWT/Internal/DefaultIDTokenValidator.swift
@@ -19,56 +19,180 @@ import CommonCrypto
 struct DefaultIDTokenValidator: IDTokenValidator {
     var issuedAtGraceInterval: TimeInterval = 300
     
+    ///  This represents the validation checks using to  validate the claim
+    enum Intent {
+        case validateIssuer(token: JWT, issuer: URL)
+        case validateAudience(token: JWT, clientId: String)
+        case validateScheme(token: JWT)
+        case validateAlgorithm(token: JWT)
+        case validateNonce(token: JWT, context: IDTokenValidatorContext?)
+        case validateIssuedAtTime(token: JWT)
+        case validateExpirationTime(token: JWT)
+        case validateAuthTime(token: JWT, context: IDTokenValidatorContext?)
+        case validateMaxAge(token: JWT, context: IDTokenValidatorContext?)
+        case validateSubject(token: JWT)
+    }
+
     func validate(token: JWT, issuer: URL, clientId: String, context: IDTokenValidatorContext?) throws {
+        let validationIntent: [Intent] = [
+            .validateIssuer(token: token, issuer: issuer),
+            .validateAudience(token: token, clientId: clientId),
+            .validateScheme(token: token),
+            .validateAlgorithm(token: token),
+            .validateExpirationTime(token: token),
+            .validateIssuedAtTime(token: token),
+            .validateNonce(token: token, context: context),
+            .validateAuthTime(token: token, context: context),
+            .validateMaxAge(token: token, context: context),
+            .validateSubject(token: token)
+        ]
+        return try self.validate(using: validationIntent)
+    }
+}
+
+extension DefaultIDTokenValidator {
+    /// Validates the claims in the given token, using the supplied intent
+    /// - Parameter intent: The validation check requirement
+    func validate(using intent: [Intent]) throws {
+         try intent.forEach({ validationIntent in
+            switch validationIntent {
+            case .validateAudience(token: let token, clientId: let clientID):
+                try self.validateAudience(using: token, clientId: clientID)
+            case .validateIssuer(token: let token, issuer: let issuer):
+                try self.validateIssuer(using: token, issuer: issuer)
+            case .validateScheme(token: let token):
+                try self.validateScheme(using: token)
+            case .validateAlgorithm(token: let token):
+                try self.validateAlgorithm(using: token)
+            case .validateNonce(token: let token, context: let context):
+                try self.validateNonce(using: token, context: context)
+            case .validateIssuedAtTime(token: let token):
+                try self.validateIssuedAtTime(using: token)
+            case .validateExpirationTime(token: let token):
+                try self.validateExpirationTime(using: token)
+            case .validateAuthTime(token: let token, context: let context):
+                try self.validateAuthTime(using: token, context: context)
+            case .validateMaxAge(token: let token, context: let context):
+                try self.validateMaxAge(using: token, context: context)
+            case .validateSubject(token: let token):
+                try self.validateSubject(using: token)
+            }
+        })
+    }
+    
+    /// Validate the Issuer claim of the token
+    /// - Parameters:
+    ///   - token: JWT token, providing access to its payload contents.
+    ///   - issuer: The supplied issuer
+    private func validateIssuer(using token: JWT, issuer: URL) throws {
         guard let tokenIssuerString = token.issuer,
-              let tokenIssuer = URL(string: tokenIssuerString)
+              let tokenIssuer = URL(string: tokenIssuerString),
+              tokenIssuer.absoluteString == issuer.absoluteString
         else {
             throw JWTError.invalidIssuer
         }
-
+        
         guard tokenIssuer.absoluteString == issuer.absoluteString
         else {
             throw JWTError.invalidIssuer
         }
-
+    }
+    
+    /// Validate the Audience claim
+    /// - Parameters:
+    ///   - token: JWT token, providing access to its payload contents.
+    ///   - clientId: The unique client ID
+    private func validateAudience(using token: JWT, clientId: String) throws {
         guard token[.audience] == clientId else {
             throw JWTError.invalidAudience
+        }
+    }
+    
+    /// Validate the Issuer scheme  of the URL.
+    /// - Parameter token: JWT token, providing access to its payload contents.
+    private func validateScheme(using token: JWT) throws {
+        guard let tokenIssuerString = token.issuer,
+            let tokenIssuer = URL(string: tokenIssuerString)
+        else {
+            throw JWTError.invalidIssuer
         }
 
         guard tokenIssuer.scheme == "https" else {
             throw JWTError.issuerRequiresHTTPS
         }
-        
+    }
+    
+    /// Validate rhe signing algorithm used to sign this JWT token.
+    /// - Parameter token: JWT token, providing access to its payload contents.
+    private func validateAlgorithm(using token: JWT) throws {
         guard token.header.algorithm == .rs256 else {
             throw JWTError.unsupportedAlgorithm(token.header.algorithm)
         }
-        
-        guard let expirationTime = token.expirationTime,
-              expirationTime > Date.nowCoordinated else {
-            throw JWTError.expired
+    }
+    
+    /// Validate the nonce claim
+    /// - Parameters:
+    ///   - token: JWT token, providing access to its payload contents.
+    private func validateNonce(using token: JWT, context: IDTokenValidatorContext?) throws {
+        guard token["nonce"] == context?.nonce else {
+            throw JWTError.nonceMismatch
         }
-        
+    }
+    
+    /// Validate the date this token was issued. iat claim
+    /// - Parameter token: JWT token, providing access to its payload contents.
+    private func validateIssuedAtTime(using token: JWT) throws {
         guard let issuedAt = token.issuedAt,
               abs(issuedAt.timeIntervalSince(Date.nowCoordinated)) <= issuedAtGraceInterval
         else {
             throw JWTError.issuedAtTimeExceedsGraceInterval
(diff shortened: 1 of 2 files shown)


Sources/AuthFoundation/JWT/Internal/DefaultIDTokenValidator.swift
```diff
@@ -19,56 +19,180 @@ import CommonCrypto
 struct DefaultIDTokenValidator: IDTokenValidator {
     var issuedAtGraceInterval: TimeInterval = 300
     
+    ///  This represents the validation checks using to  validate the claim
+    enum Intent {
```
mikenachbaur-okta, line 23: Supplying the token, context, issuer, etc here seems to complicate the enum without adding much value, because these arguments are passed to private functions within the DefaultIDTokenValidator in the end. 

It also runs the risk of unrelated JWT tokens being passed to the same validator call. 

Perhaps a better approach would be to keep the original validator method signature, and add a `tolenValidation` property (or something along those lines) to the Context object. The validation can take those settings and can use that internally to perform validation steps.

You could also supply a static property on the Intent enum called `all`, which not only includes the full set of validation steps, but is the default if none is supplied. 

This is because the developer doesn't directly invoke the validate function, but is rather automatically called after a token is exchanged. So the developer would need to pass these options at the time they initiate an authentication flow so as to control the behavior later on downstream. 

An alternative might be to make this a property on the token validation itself, that can be assigned at app launch. But that then means different tokens can't be validated with different verification steps (which I could imagine would be important to some use-cases).

mikenachbaur-okta (changes requested):
I left a lengthy comment,  it I really do like the direction you've headed. Let me know what you think, and we can chat about this in more detail. 

We also should loop in @jaynewstrom-okta since this is something we'd want parity on.

Sources/AuthFoundation/JWT/Internal/DefaultIDTokenValidator.swift
emmanuelogunmuyiwa-okta: Thanks alot @mikenachbaur-okta . There's alot of good callout there. 
- I'll make changes to the Intent, we definitely do not need to supply the token, context etc to the intent. 
- I also think we can use an OptionSet here as well to define the order.

Sources/AuthFoundation/Token Management/IDTokenValidator.swift
```diff
@@ -36,4 +36,7 @@ public protocol IDTokenValidatorContext {
     
     /// The maximum age the token should support when authenticating.
     var maxAge: TimeInterval? { get }
+    
+    /// The possible option using  in validating a claim
+    var validationOptions: ValidationOption? { get }
```
JayNewstrom, line 41: All of these options are required by the spec. https://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation

If someone wants to break the spec, I don't mind forcing pain on them (by having them reimplement this class, likely via copy/paste and modifying the lines they want changed).

Sources/AuthFoundation/JWT/Internal/DefaultIDTokenValidator.swift
emmanuelogunmuyiwa-okta: I've updated this.

Sources/AuthFoundation/Token Management/IDTokenValidator.swift
emmanuelogunmuyiwa-okta: I agree, this is definitely something we need to think about. I think in this case the spec defines the order/manner for the validation. _(Which brings a valid question of why make it modifiable by a client.)_ . My opinion is we can guarantee the order at which the validation occurs and provide the flexibility on the type of validation that occurs _(if someone has a use case that requires skipping a check)_

mikenachbaur-okta: All good points. I was thinking about future-proofing for the real-world issues that some developers face, but at the outset we shouldn't optimize for bypassing security checks. If anything, the original intent of this PR (to simplify the readability and testability of validation checks) still stands. If we need to make this customizable in the future, this gives us a path forward to do that without a lot of effort.

Sources/AuthFoundation/JWT/Internal/DefaultIDTokenValidator.swift
```diff
@@ -16,59 +16,212 @@ import Foundation
 import CommonCrypto
 #endif
 
+/// This represents the possible option using in validating a claim
+public struct ValidationOption: OptionSet  {
```
mikenachbaur-okta, line 20: This seems a bit verbose to me. It feels like a lot more code than the original implementation.

What's wrong with using CaseIterable?

```swift
struct DefaultIDTokenValidator: IDTokenValidator {
    var checks: [ValidationCheck] = ValidationCheck.allCases
    
    enum ValidationCheck: CaseIterable {
        case issuer, audience, scheme, algorithm, expirationTime, issuedAtTime, nonce, authTime, maxAge, subject
    }

   ...

       func validate(token: JWT, issuer: URL, clientId: String, context: IDTokenValidatorContext?) throws {
        for check in checks {
          // Put existing checks here
        }
```

Sources/AuthFoundation/JWT/Internal/DefaultIDTokenValidator.swift
```diff
@@ -16,59 +16,212 @@ import Foundation
...
+    /// Validate the subject of the resource, if available.
+    static let validateSubject = ValidationOption(rawValue: 1 << 9)
+}
+
 struct DefaultIDTokenValidator: IDTokenValidator {
     var issuedAtGraceInterval: TimeInterval = 300
-    
+        
     func validate(token: JWT, issuer: URL, clientId: String, context: IDTokenValidatorContext?) throws {
+        let validationOptions = context?.validationOptions ?? .all
```
mikenachbaur-okta, line 67: If we're not making the validation options public, we don't need to put it into the context anymore.  That's something that can be added later, if necessary.

Sources/AuthFoundation/JWT/Internal/DefaultIDTokenValidator.swift
```diff
@@ -16,59 +16,212 @@ import Foundation
...
+
```
mikenachbaur-okta, line 68: I don't think we need to break all of these into separate functions. If anything, it would just add to the symbol table overhead.

I'm thinking something like this:

```swift
        for check in checks {
            switch check {
            case .issuer:
                guard tokenIssuer.absoluteString == issuer.absoluteString
                else {
                    throw JWTError.invalidIssuer
                }
                
            case .audience:
                guard token[.audience] == clientId else {
                    throw JWTError.invalidAudience
                }
   ...
```

The nice thing is the diff for this commit will be minimal, since the code will be laid out mostly the same, just a few lines of indentation and a switch / case block.

Sources/AuthFoundation/JWT/Internal/DefaultIDTokenValidator.swift
```diff
@@ -16,59 +16,212 @@ import Foundation
...
+        }
+    }
+}
+
+extension DefaultIDTokenValidator {
+    /// Validate the Issuer claim of the token
+    /// - Parameters:
+    ///   - token: JWT token, providing access to its payload contents.
+    ///   - issuer: The supplied issuer
+    private func validateIssuer(using token: JWT, issuer: URL) throws {
```
mikenachbaur-okta, line 116: We don't need to document internal / private functions.

Sources/AuthFoundation/Token Management/IDTokenValidator.swift
mikenachbaur-okta, line 41: We don't need this if we're not making it public anymore.

mikenachbaur-okta (changes requested):

Sources/AuthFoundation/JWT/Internal/DefaultIDTokenValidator.swift
```diff
@@ -18,61 +18,72 @@ import CommonCrypto
...
+            case .expirationTime:
+                guard let expirationTime = token.expirationTime,
+                      expirationTime > Date.nowCoordinated else {
+                    throw JWTError.expired
+                }
+            case .nonce:
+                guard token["nonce"] == context?.nonce else {
+                    throw JWTError.nonceMismatch
+                }
+            case .issuedAtTime, .authTime, .maxAge:
```
mikenachbaur-okta, line 63: We might want to break these out into separate cases, so they can be tested discretely.

Also, I'm not entirely certain if we need a separate "authTime" check.  There are two separate exceptions that may be thrown, but `.invalidAuthenticationTime` and `.exceedsMaxAge` are just different errors that can occur while testing `maxAge`

Sources/AuthFoundation/JWT/Internal/DefaultIDTokenValidator.swift
```diff
@@ -18,61 +18,72 @@ import CommonCrypto
 
 struct DefaultIDTokenValidator: IDTokenValidator {
     var issuedAtGraceInterval: TimeInterval = 300
+    var checks: [ValidationCheck] = ValidationCheck.allCases
+    
+    enum ValidationCheck: CaseIterable {
+        case issuer, audience, scheme, algorithm, expirationTime, issuedAtTime, nonce, authTime, maxAge, subject
+    }
     
     func validate(token: JWT, issuer: URL, clientId: String, context: IDTokenValidatorContext?) throws {
```
mikenachbaur-okta, line 27: You might want to change the `.swiftlint.yaml` to increase the cyclomatic_complexity threshold.  Lets set it to 25 or so.  I'm not a fan of that particular lint check, to be honest...

emmanuelogunmuyiwa-okta: Good point, I've updated this.  I did set a warning and error configuration.

Sources/AuthFoundation/JWT/Internal/DefaultIDTokenValidator.swift
```diff
@@ -18,61 +18,72 @@ import CommonCrypto
...
+                guard tokenIssuer.scheme == "https" else {
+                    throw JWTError.issuerRequiresHTTPS
+                }
+            case .algorithm:
+                guard token.header.algorithm == .rs256 else {
+                    throw JWTError.unsupportedAlgorithm(token.header.algorithm)
+                }
+            case .expirationTime:
+                guard let expirationTime = token.expirationTime,
+                      expirationTime > Date.nowCoordinated else {
```
mikenachbaur-okta, line 56: Can you put this `else` on a new line?

The style guide I've been working with is inline `guard else` can be fine for single expressions, but for multiple expressions, we put the `else` on its own line.  This is because otherwise the expression, and the code, can be hard to distinguish from one another.

Sources/AuthFoundation/JWT/Internal/DefaultIDTokenValidator.swift
```diff
@@ -18,61 +18,72 @@ import CommonCrypto
...
-            }
-            
-            let elapsedTime = issuedAt.timeIntervalSince(authTime)
-            guard elapsedTime > 0 && elapsedTime <= maxAge else {
-                throw JWTError.exceedsMaxAge
+        for check in checks {
+            switch check {
+            case .issuer:
+                guard tokenIssuer.absoluteString == issuer.absoluteString
+                else {
```
mikenachbaur-okta, line 39: You can choose to put this `else` on the same line as `guard` if you want. But, with our style, we optimize for readability over purity.

Sources/AuthFoundation/JWT/Internal/DefaultIDTokenValidator.swift
```diff
@@ -18,61 +18,72 @@ import CommonCrypto
...
+                    throw JWTError.nonceMismatch
+                }
+            case .issuedAtTime:
+                guard let issuedAt = token.issuedAt,
+                      abs(issuedAt.timeIntervalSince(Date.nowCoordinated)) <= issuedAtGraceInterval
+                else {
+                    throw JWTError.issuedAtTimeExceedsGraceInterval
+                }
+            case .maxAge:
+                if let maxAge = context?.maxAge, let issuedAt = token.issuedAt {
```
mikenachbaur-okta, line 70: Can you break each condition out onto its own line?
```suggestion
                if let maxAge = context?.maxAge,
                    let issuedAt = token.issuedAt
                {
```

mikenachbaur-okta (commented):
Minor formatting comments, but looks good!

Sources/AuthFoundation/JWT/Internal/DefaultIDTokenValidator.swift
emmanuelogunmuyiwa-okta: The sounds good. Maybe we can add this to the linter, so its not something we manually check for down the line. I'll look into it later.

mikenachbaur-okta (approved):
This looks great, thanks!

PR #86, opened 2022-07-14
Use openIdConfiguration issuer when verifying ID token  issuer claim

emmanuelogunmuyiwa-okta:
* To support the use of vanity domains, it’s important to use the issuer URL from the opened-configuration when performing ID Token validation.

### DIFF
--- a/Sources/AuthFoundation/OAuth2/OAuth2Client.swift
+++ b/Sources/AuthFoundation/OAuth2/OAuth2Client.swift
@@ -422,30 +422,38 @@ public final class OAuth2Client {
             }
         }
         
-        var configuration = openIdConfiguration
-        if configuration == nil {
-            group.enter()
-            self.openIdConfiguration { result in
-                defer { group.leave() }
-                if case let .success(response) = result {
-                    configuration = response
-                }
+        // Exchange the token
+        request.send(to: self) { result in
+            // Wait for the JWKS keys, if necessary
+            group.notify(queue: DispatchQueue.global()) {
+                // Perform idToken/accessToken validation
+                self.validateToken(
+                    request: request,
+                    keySet: keySet,
+                    oauthTokenResponse: result,
+                    completion: completion
+                )
             }
         }
+    }
+    
+    private func validateToken<T: OAuth2TokenRequest>(
+        request: T,
+        keySet: JWKS?,
+        oauthTokenResponse: Result<APIResponse<Token>, APIClientError>,
+        completion: @escaping (Result<APIResponse<Token>, APIClientError>) -> Void
+    ) {
+        guard case let .success(response) = oauthTokenResponse else {
+            completion(oauthTokenResponse)
+            return
+        }
         
-        group.notify(queue: DispatchQueue.global()) {
-            // Exchange the token
-            request.send(to: self) { result in
-                guard case let .success(response) = result else {
-                    completion(result)
-                    return
-                }
-        
-                guard let configuration = configuration else {
-                    completion(.failure(.validation(error: JWTError.invalidIssuer)))
-                    return
-                }
-                
+        // Retrieves the org's OpenID configuration
+        self.openIdConfiguration { result in
+            switch result {
+            case .failure(let error):
+                completion(.failure(.serverError(error)))
+            case .success(let configuration):
                 do {
                     try response.result.validate(
                         using: self,
@@ -457,30 +465,26 @@ public final class OAuth2Client {
                     return
                 }
                 
-                // Wait for the JWKS keys, if necessary
-                group.notify(queue: DispatchQueue.global()) {
-                    guard let idToken = response.result.idToken else {
-                        completion(result)
-                        return
-                    }
-                    
-                    guard let keySet = keySet else {
-                        completion(.failure(.validation(error: JWTError.invalidKey)))
-                        return
-                    }
-                        
-                    do {
-                        if try idToken.validate(using: keySet) == false {
-                            completion(.failure(.validation(error: JWTError.signatureInvalid)))
-                            return
-                        }
-                    } catch {
-                        completion(.failure(.validation(error: error)))
-                        return
-                    }
-                    
-                    completion(result)
+                guard let idToken = response.result.idToken else {
+                    completion(oauthTokenResponse)
+                    return
                 }
+                
+                guard let keySet = keySet else {
+                    completion(.failure(.validation(error: JWTError.invalidKey)))
+                    return
+                }
+                
+                do {
+                    if try idToken.validate(using: keySet) == false {
+                        completion(.failure(.validation(error: JWTError.signatureInvalid)))
+                        return
+                    }
+                } catch {
+                    completion(.failure(.validation(error: error)))
+                    return
+                }
+                completion(oauthTokenResponse)
             }
         }
     }
--- a/Sources/AuthFoundation/OAuth2/OAuth2Client.swift
+++ b/Sources/AuthFoundation/OAuth2/OAuth2Client.swift
@@ -424,7 +424,7 @@ public final class OAuth2Client {
 
         openIdConfiguration { result in
             switch result {
-            case .success(let config):
+            case .success(let configuration):
                 // Exchange the token
                 request.send(to: self) { result in
                     guard case let .success(response) = result else {
@@ -433,7 +433,7 @@ public final class OAuth2Client {
                     }
                     
                     do {
-                        try response.result.validate(using: self, issuer: config.issuer, with: request as? IDTokenValidatorContext)
+                        try response.result.validate(using: self, issuer: configuration.issuer, with: request as? IDTokenValidatorContext)
                     } catch {
                         completion(.failure(.validation(error: error)))
                         return
--- a/Sources/AuthFoundation/OAuth2/OAuth2Client.swift
+++ b/Sources/AuthFoundation/OAuth2/OAuth2Client.swift
@@ -450,11 +450,9 @@ public final class OAuth2Client {
             switch result {
             case .failure(let error):
                 completion(.failure(.serverError(error)))
-            case .success(let configuration):
+            case .success:
                 do {
-                    try response.result.validate(using: self,
-                                                 issuer: configuration.issuer,
-                                                 with: request as? IDTokenValidatorContext)
+                    try response.result.validate(using: self, with: request as? IDTokenValidatorContext)
                 } catch {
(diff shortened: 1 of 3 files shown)


Tests/OktaOAuth2Tests/OAuth2ClientTests.swift
```diff
@@ -81,4 +81,45 @@ final class OAuth2ClientTests: XCTestCase {
         XCTAssertNotNil(token?.idToken)
         XCTAssertEqual(token?.scope, "openid profile offline_access")
     }
+    
+
+    func testExchangeFailed() throws {
+        let pkce = PKCE()
+        client = OAuth2Client(baseURL: issuer, clientId: "theClientId", scopes: "openid profile offline_access", session: urlSession)
```
mikenachbaur-okta, line 88: Can you indent this across multiple lines, for readability?

Sources/AuthFoundation/OAuth2/OAuth2Client.swift
```diff
@@ -421,45 +421,51 @@ public final class OAuth2Client {
...
-                    completion(result)
-                    return
-                }
-                
-                guard let keySet = keySet else {
-                    completion(.failure(.validation(error: JWTError.invalidKey)))
-                    return
-                }
+
+        openIdConfiguration { result in
```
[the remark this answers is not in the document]
emmanuelogunmuyiwa-okta: Good point. This had me thinking as well, on the off chance openId configuration isn't cached then that could be an indication that  something has gone wrong with the flow and we should throw an error. since the flow is triggered  at the `start` or at `resume` and they both call to `openIdConfiguration()`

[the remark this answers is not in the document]
mikenachbaur-okta: That's true, though we may change the model for how caching of openIdConfiguration works in the future. So it's better for us to protect against it, and ensure the operations that _could_ be async don't block the token exchange.

For more context, we fetch the JWKS keys in parallel with the token exchange; they don't require one-another, but we need both to complete in order to validate the returned tokens. As a result, it makes sense to use that same parallelization system to fetch configuration as well.

Sources/AuthFoundation/OAuth2/OAuth2Client.swift
```diff
@@ -429,14 +429,26 @@ public final class OAuth2Client {
...
             }
             
-            // Perform idToken/accessToken validation
-            do {
-                try response.result.validate(using: self, with: request as? IDTokenValidatorContext)
-            } catch {
-                completion(.failure(.validation(error: error)))
-                return
+            group.enter()
+            self.openIdConfiguration { result in
```
mikenachbaur-okta, line 433: This is still serializing the requests, since you're kicking off the request for openIdConfiguration within the `request.send` completion block.  Can you use the fetch for the keySet up on line 415 as a guide?

emmanuelogunmuyiwa-okta: Thanks for pointing me in that direction. I have updated this.

Sources/AuthFoundation/OAuth2/OAuth2Client.swift
```diff
@@ -422,56 +422,65 @@ public final class OAuth2Client {
...
-                group.leave()
             }
+        }
         
-            // Wait for the JWKS keys, if necessary
-            group.notify(queue: DispatchQueue.global()) {
-                guard let idToken = response.result.idToken else {
+        group.notify(queue: DispatchQueue.global()) {
+            // Exchange the token
+            request.send(to: self) { result in
```
mikenachbaur-okta, line 438: There's no need to use `group.notify` here, because we want the token exchange request to happen in parallel with the jwks and openIdConfiguration requests. So let this one run immediately, and use the `group.notify` you have later in this method to handle the results.

Sources/AuthFoundation/OAuth2/OAuth2Client.swift
```diff
@@ -424,31 +424,57 @@ public final class OAuth2Client {
...
+                    request: request,
+                    keySet: keySet,
+                    oauthTokenResponse: result,
+                    completion: completion
+                )
+            }
+        }
+    }
+    
+    private func validateToken<T: OAuth2TokenRequest>(
```
mikenachbaur-okta, line 440: Can we format the functions with the first argument inline, and subsequent ones indented to match?  The other functions can serve as an example.

Sources/AuthFoundation/OAuth2/OAuth2Client.swift
```diff
@@ -424,31 +424,57 @@ public final class OAuth2Client {
...
-            group.notify(queue: DispatchQueue.global()) {
+            group.notify(queue: DispatchQueue.global()) { [weak self] in
+                // Perform idToken/accessToken validation
+                self?.validateToken(
+                    request: request,
+                    keySet: keySet,
+                    oauthTokenResponse: result,
+                    completion: completion
+                )
+            }
```
mikenachbaur-okta, line 436: Can you put the first argument inline, so the other methods indent along with them?
```suggestion
                self?.validateToken(request: request,
                                    keySet: keySet,
                                    oauthTokenResponse: result,
                                    completion: completion)
            }
```
```suggestion
                self?.validateToken(
                    request: request,
                    keySet: keySet,
                    oauthTokenResponse: result,
                    completion: completion
                )
            }
```

Sources/AuthFoundation/OAuth2/OAuth2Client.swift
```diff
@@ -424,31 +424,57 @@ public final class OAuth2Client {
...
+        }
+    }
+    
+    private func validateToken<T: OAuth2TokenRequest>(
+        request: T,
+        keySet: JWKS?,
+        oauthTokenResponse: Result<APIResponse<Token>, APIClientError>,
+        completion: @escaping (Result<APIResponse<Token>, APIClientError>) -> Void
+    ) {
```
mikenachbaur-okta, line 445: Can we format the functions with the first argument inline, and subsequent ones indented to match?  The other functions can serve as an example.
```suggestion
    private func validateToken<T: OAuth2TokenRequest>(request: T,
                                                      keySet: JWKS?,
                                                      oauthTokenResponse: Result<APIResponse<Token>, APIClientError>,
                                                      completion: @escaping (Result<APIResponse<Token>, APIClientError>) -> Void)
    {
```

Sources/AuthFoundation/OAuth2/OAuth2Client.swift
```diff
@@ -424,31 +424,57 @@ public final class OAuth2Client {
...
+        }
+        
+        // Retrieves the org's OpenID configuration
+        self.openIdConfiguration { result in
+            switch result {
+            case .failure(let error):
+                completion(.failure(.serverError(error)))
+            case .success(let configuration):
+                do {
+                    try response.result.validate(
```
mikenachbaur-okta, line 458: Similar thing here. Elsewhere in the project, this is the formatting style used, so I just want to keep it consistent.

Sources/AuthFoundation/Token Management/Token.swift
```diff
@@ -78,13 +78,14 @@ public final class Token: Codable, Equatable, Hashable, Expires {
     
     /// Validates the claims within this JWT token, to ensure it matches the given ``OAuth2Client``.
     /// - Parameter client: Client to validate the token's claims against.
-    public func validate(using client: OAuth2Client, with context: IDTokenValidatorContext?) throws {
+    /// - Parameter issuer: The client issuer URL.
+    public func validate(using client: OAuth2Client, issuer: URL, with context: IDTokenValidatorContext?) throws {
```
mikenachbaur-okta, line 82: Hmm the more I think about it, I'm not sure about asking for the issuer URL since it's an extra step, and isn't needed for the majority of use-cases.  Especially since, in the future, we'll be caching the OpenIdConfiguration.

How about we make the issuer optional, with a default that falls-back to the cached openid configuration?  What are your thoughts?

Sources/AuthFoundation/OAuth2/OAuth2Client.swift
emmanuelogunmuyiwa-okta: Sounds good.

Sources/AuthFoundation/Token Management/Token.swift
emmanuelogunmuyiwa-okta: - I was thinking about it from the perspective that for an "iss" claim to be valid, the issuer url would need to be provided to `IDTokenValidator` and would need to pass the check else we through an error. 

- If we make it optional, the default (_the cached openid configuration_)is also an optional as well. Which would require changes to the   IDTokenValidator because thats a required parameter.

Also I think the validate functionality should be strictly enforced`(we can't validate if the required parameters are not provided)` 

-From my understanding, the only  case where the  openIdConfiguration( ) is not called and `issuerUrl` is not cached is only the from the idx sdk and thats why we call the  openIdConfiguration()  in parallel if its ever nil

emmanuelogunmuyiwa-okta: The more I think about it, since the `openid configuration` is cached in memory, we can also not ask for the issuer URL here.
```
    public func validate(using client: OAuth2Client, with context: IDTokenValidatorContext?) throws {
        guard let idToken = idToken, let issuer = client.openIdConfiguration?.issuer else {
            return
        }
```

However I still have concerns about not requiring the `issuer` since it's a requirement to validate a token.

mikenachbaur-okta: So if we unwrap the option here, and throw an appropriate error instead of silently returning, that is probably acceptable.

Sources/AuthFoundation/OAuth2/OAuth2Client.swift
```diff
@@ -424,31 +424,50 @@ public final class OAuth2Client {
...
-            do {
-                try response.result.validate(using: self, with: request as? IDTokenValidatorContext)
-            } catch {
-                completion(.failure(.validation(error: error)))
-                return
-            }
-            
             // Wait for the JWKS keys, if necessary
-            group.notify(queue: DispatchQueue.global()) {
+            group.notify(queue: DispatchQueue.global()) { [weak self] in
```
mikenachbaur-okta, line 428: Does this need to be a weak reference? It doesn't look like it will create a retain cycle.

mikenachbaur-okta (approved):
Great work!

PR #89, opened 2022-07-20
Relax the values an OpenIdConfiguration requires, and enable custom claim names

mikenachbaur-okta:
This was discovered when trying other authorization servers, such as AppleId.

### DIFF
--- a/Sources/AuthFoundation/JWT/Claim.swift
+++ b/Sources/AuthFoundation/JWT/Claim.swift
@@ -13,150 +13,225 @@
 import Foundation
 
 /// List of registered and public claims.
-public enum Claim: String, Codable {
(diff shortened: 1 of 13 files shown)


emmanuelogunmuyiwa-okta:
Looks good! just a minor comment, I think in the future we can improve the test coverages by including tests for the not so happy path where errors are thrown. eg in the _SessionLogoutFlow_, _OAuth2Client_

mikenachbaur-okta:
I agree 💯! We definitely have some test debt in the non-happy-path scenarios (which, lets be honest, are the more important cases to cover).  We'll continue to pay that debt down over the coming weeks.

emmanuelogunmuyiwa-okta (approved):

PR #90, opened 2022-07-22
Introduce`revoke(.all)` and associated documentation

mikenachbaur-okta:
This clears up a gap in functionality between this SDK and `okta-oidc-ios`, in that the legacy SDK had the capability of revoking all tokens through one function.  This also adds documentation to clarify how to properly sign a user out.


Sources/AuthFoundation/OAuth2/OAuth2Client.swift
```diff
@@ -272,7 +272,16 @@ public final class OAuth2Client {
     ///   - type: Type of token to revoke.
     ///   - completion: Completion block to invoke once complete.
     public func revoke(_ token: Token, type: Token.RevokeType, completion: @escaping (Result<Void, OAuth2Error>) -> Void) {
-        let tokenType = type.tokenType
+        if type == .all {
+            revokeAll(token, completion: completion)
+            return
+        }
+        
+        guard let tokenType = type.tokenType else {
```
emmanuelogunmuyiwa-okta, line 280: In the case where `type.tokenType == nil` it would also be map to `type == .all `. I think we can just call the             `revokeAll(token, completion: completion)` if the unwrapping fails. However, I was wondering why the first check `if type == .all. ` exists if the optional unwrap also covers it.  

Although if the first is true, we would never get to the optional unwrap.

Sources/AuthFoundation/OAuth2/OAuth2Client.swift
```diff
@@ -455,6 +464,43 @@ public final class OAuth2Client {
         }
     }
     
+    private func revokeAll(_ token: Token, completion: @escaping (Result<Void, OAuth2Error>) -> Void) {
+        let types: [Token.RevokeType] = [.accessToken, .refreshToken, .deviceSecret]
+        
+        var errors = [OAuth2Error]()
```
emmanuelogunmuyiwa-okta, line 470: Minor comment: Maybe we can remove this mutability by using reduce into.  I was thinking something like this. 
 
```

        let revokeTypes = types.map({ type -> Token.RevokeType? in
            if let revokeType = type.tokenType, token.token(of: revokeType) != nil {
                return type
            }
            return nil
        })
   
        
        let errors = revokeTypes.compactMap({ $0 })
            .reduce(into: [OAuth2Error](), { newErrors, type in
                group.enter()
```

Sources/AuthFoundation/OAuth2/OAuth2Client.swift
mikenachbaur-okta: I was trying to avoid force-unwrapping `type.tokenType` since its become optional now. There's no guarantee, in the future, that other types could be introduced _other_ than `.all` which may also not return a `tokenType`. I think I could make the `if` on the first line of this function a bit cleaner though.

Sources/AuthFoundation/OAuth2/OAuth2Client.swift
mikenachbaur-okta: Reduce is synchronous, so I'd need to introduce some type of asynchronous `reduce` implementation to ensure the results could be accumulated. Additionally, I'd still want these requests to occur concurrently, which would be difficult to reconcile with `reduce`.

So at this time, without introducing additional dependencies, or requiring Swift Concurrency support, this is probably the cleanest way to accomplish this.

Sources/AuthFoundation/OAuth2/OAuth2Client.swift
emmanuelogunmuyiwa-okta: That sounds good. I see your point here.

Sources/AuthFoundation/OAuth2/OAuth2Client.swift
emmanuelogunmuyiwa-okta: Good point, I don't think that would be necessary at all.

emmanuelogunmuyiwa-okta (approved):

PR #91, opened 2022-08-08
Supply client_id to Introspect request call

emmanuelogunmuyiwa-okta:
- Include the client_id as a query parameter when calling the /introspect endpoint. 
- do not pass the Authorization header in the request
- Link to swagger doc:  https://developer.okta.com/docs/reference/api/oidc/#introspect

### DIFF
--- a/Sources/AuthFoundation/Requests/Token+Requests.swift
+++ b/Sources/AuthFoundation/Requests/Token+Requests.swift
@@ -82,10 +82,12 @@ extension Token.IntrospectRequest: OAuth2APIRequest, APIRequestBody {
     var httpMethod: APIRequestMethod { .post }
     var contentType: APIContentType? { .formEncoded }
     var acceptsType: APIContentType? { .json }
-    var authorization: APIAuthorization? { token }
+    var authorization: APIAuthorization? { nil }
+    private var clientId: String? { token.idToken?[.clientId] }
     var bodyParameters: [String: Any]? {
         [
             "token": (token.token(of: type) ?? "") as String,
+            "client_id": token.context.configuration.clientId,
             "token_type_hint": type.rawValue
         ]
     }
--- a/Sources/AuthFoundation/Requests/Token+Requests.swift
+++ b/Sources/AuthFoundation/Requests/Token+Requests.swift
@@ -83,7 +83,6 @@ extension Token.IntrospectRequest: OAuth2APIRequest, APIRequestBody {
     var contentType: APIContentType? { .formEncoded }
     var acceptsType: APIContentType? { .json }
     var authorization: APIAuthorization? { nil }
-    private var clientId: String? { token.idToken?[.clientId] }
     var bodyParameters: [String: Any]? {
         [
             "token": (token.token(of: type) ?? "") as String,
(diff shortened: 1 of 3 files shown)


Sources/AuthFoundation/Requests/Token+Requests.swift
```diff
@@ -82,10 +82,12 @@ extension Token.IntrospectRequest: OAuth2APIRequest, APIRequestBody {
     var httpMethod: APIRequestMethod { .post }
     var contentType: APIContentType? { .formEncoded }
     var acceptsType: APIContentType? { .json }
-    var authorization: APIAuthorization? { token }
+    var authorization: APIAuthorization? { nil }
+    private var clientId: String? { token.idToken?[.clientId] }
     var bodyParameters: [String: Any]? {
         [
             "token": (token.token(of: type) ?? "") as String,
+            "client_id": token.context.configuration.clientId,
```
JayNewstrom, line 90: Can you add a test that expects this param to exist? 

Here's one I did in kotlin: https://github.com/okta/okta-mobile-kotlin/blob/master/auth-foundation/src/test/java/com/okta/authfoundation/client/OidcClientTest.kt#L211

Sources/AuthFoundation/Requests/Token+Requests.swift
```diff
@@ -82,10 +82,12 @@ extension Token.IntrospectRequest: OAuth2APIRequest, APIRequestBody {
     var httpMethod: APIRequestMethod { .post }
     var contentType: APIContentType? { .formEncoded }
     var acceptsType: APIContentType? { .json }
-    var authorization: APIAuthorization? { token }
+    var authorization: APIAuthorization? { nil }
+    private var clientId: String? { token.idToken?[.clientId] }
```
mikenachbaur-okta, line 86: You can delete this line, since it's not used.

Sources/AuthFoundation/Requests/Token+Requests.swift
emmanuelogunmuyiwa-okta: Good call out @JayNewstrom . I have include test for that case

Sources/AuthFoundation/Requests/Token+Requests.swift
emmanuelogunmuyiwa-okta: I must've missed this, thanks. I have updated this.

Tests/AuthFoundationTests/OAuth2ClientTests.swift
```diff
@@ -143,7 +143,91 @@ final class OAuth2ClientTests: XCTestCase {
...
+        let openIdConfiguration = try OpenIdConfiguration.jsonDecoder.decode(
+            OpenIdConfiguration.self,
+            from: try data(from: .module,
+                           for: "openid-configuration",
+                           in: "MockResponses"))
+        let request = try Token.IntrospectRequest(openIdConfiguration: openIdConfiguration,
+                                                  token: token,
+                                                  type: .accessToken)
+        XCTAssertNil(request.authorization)
+        XCTAssertEqual((request.bodyParameters?["client_id"] as? String), token.context.configuration.clientId)
```
JayNewstrom, line 156: I'd suggest asserting something concrete, rather than something like `token.context.configuration.clientId`.

It's harder to know what's expected when it's coming from somewhere else, and it's also easy to break your expectations without knowing it.

emmanuelogunmuyiwa-okta: Good callout!. I've updated this.

JayNewstrom (approved):

mikenachbaur-okta (approved):

PR #93, opened 2022-09-09
Add support for retrying 429 errors

emmanuelogunmuyiwa-okta:


Sources/AuthFoundation/Network/APIClient.swift
```diff
@@ -116,14 +157,60 @@ extension APIClient {
...
+                    case .doNotRetry: break
+                    case .retry(let maximumRetryCount):
+                        let retryState: APIRetry.State
+                        if let state = state {
+                            retryState = state.nextState()
+                        } else {
+                            guard let requestIdHeader = requestIdHeader,
+                                  let requestId = httpResponse.allHeaderFields[requestIdHeader] as? String
+                            else {
+                                throw APIClientError.noRequestId
```
JayNewstrom, line 190: I don't think we want to throw an exception in this case. We're mostly doing this for traceability, it's not a strict requirement for it to work (to my knowledge)

emmanuelogunmuyiwa-okta: The requestId is added to the retry requests` X-Okta-Retry-For` which is necessary from my understanding. without throwing an exception in cases where it's missing from my original request, we would not be adding that to the retry request header

JayNewstrom: Right. I'm suggesting the `X-Okta-Retry-For` is optional, and not including it should not be considered an error.

emmanuelogunmuyiwa-okta: Okay. I think that's a good suggestion. Seems like that would be consistent with the donet implementation as well. I'll make the changes.

Sources/AuthFoundation/Network/APIClient.swift
```diff
@@ -116,14 +157,59 @@ extension APIClient {
...
+                        return
+                    }
+                    fallthrough
+                default:
+                    if let error = error(from: data) ?? context?.error(from: data) {
+                        throw APIClientError.serverError(error)
+                    } else {
+                        throw APIClientError.statusCode(httpResponse.statusCode)
+                    }
+                }
```
rajdeepnanua-okta, line 212: Shouldn't this be handled by existing code already? I'm not sure why more error handling for non-429 cases needs to be added here.

Sources/AuthFoundation/Network/APIResponse.swift
```diff
@@ -29,39 +29,47 @@ public struct APIResponse<T: Decodable>: Decodable {
...
+        self.limit = rateLimit
+        self.remaining = remaining
+        self.reset = reset
+       
+        var date: Date? = nil
+        if let dateString = httpHeaders["Date"] as? String {
+            date = httpDateFormatter.date(from: dateString)
         }
+        
+        self.delay = reset - (date ?? Date()).timeIntervalSince1970
```
rajdeepnanua-okta, line 71: I noticed that there isn't a minDelay. I've added that in the kotlin implementation at https://github.com/okta/okta-mobile-kotlin/pull/215. I prefer both implementations to have a similar feature set. Also, I'm not too familiar with swift, but what happens in the case when either of "date" header or "x-rate-limit-reset" header is missing or invalid?

rajdeepnanua-okta (commented):
I noticed that this implementation doesn't have minDelay, but that is part of the kotlin changes, and the design document. I would also suggest adding tests that check the behavior with invalid "Date" and "X-Rate-Limit-Reset". Also, a test for checking that the SDK is waiting the correct amount of time before retrying should be added.

Sources/AuthFoundation/Network/APIClient.swift
```diff
@@ -134,12 +220,21 @@ extension APIClient {
             }
         }.resume()
     }
+    
+    private func addRetryHeadersToRequest(state: APIRetry.State) -> URLRequest {
+        var request = state.originalRequest
+        if let requestId = state.requestId {
+            request.allHTTPHeaderFields?.updateValue(requestId, forKey: "X-Okta-Retry-For")
```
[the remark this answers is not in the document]
emmanuelogunmuyiwa-okta: Since the header is optional, we won't be throwing/catching any exception on the client side I don't think checking the X-Okta-Retry-For would add any benefit. I did add test for the `X-Okta-Retry-Count` header because the count is required.

Sources/AuthFoundation/Network/APIResponse.swift
emmanuelogunmuyiwa-okta: After having a conversation with @mikenachbaur-okta . The idea is to make this as server driven as possible, hence we did not include a random minimum delay. I'm open to suggestions though

Sources/AuthFoundation/Network/APIClient.swift
emmanuelogunmuyiwa-okta: This is the default fallthrough case.
```
-  if success code ... we handle accordingly 
- If 429 error we handle accordingly. 
- else : check the default case. 
```

Sources/AuthFoundation/Network/APIClient.swift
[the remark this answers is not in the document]
rajdeepnanua-okta: I think it's still worthwhile to have a test just for checking that we're adding the header correctly if it's present. It doesn't have to be in every test, like with X-Okta-Retry-Count

Tests/AuthFoundationTests/ApiRetryTests.swift
```diff
@@ -0,0 +1,281 @@
...
+// WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//
+// See the License for the specific language governing permissions and limitations under the License.
+//
+
+import XCTest
+@testable import AuthFoundation
+@testable import TestCommon
+
+class ApiRetryTests: XCTestCase {
```
mikenachbaur-okta, line 17: Since many of these tests use the same expectation (e.g. identical response headers and status code) you might be able to simplify this by reusing the same expectation.  You can either create a function to create this expectation `n` number of times, or you can rearrange some of the logic in `URLSessionMock.swift` to allow you to more simply create a `URLSessionMock.Call` object.  For example, in `URLSessionMock.expect(_:data:statusCode:contentType:error:)`, you can move the creation of the response to a convenience initializer on the `Call` type, allowing you to create a `let` value for the call.

Either way, I think this code can be cleaned up a bit.

Sources/AuthFoundation/Network/APIResponse.swift
rajdeepnanua-okta: When falling back to local date, it might be possible for ```date``` to be greater than ```reset```. In a case like that, I think it makes sense to have a minDelay. Thoughts?

mikenachbaur-okta: I feel that not getting a delay from the server, or being unable to calculate a proper delay, should be a error case. Instead of guessing a minimum delay, we should just fall through to the error case.

We have another ticket to provide out-of-the-box time sync/coordination support, so client clock skew shouldn't be an issue once we do that.

Sources/AuthFoundation/Network/APIResponse.swift
```diff
@@ -29,39 +29,47 @@ public struct APIResponse<T: Decodable>: Decodable {
     public let links: [Link: URL]
     
     /// Information about the current rate limit.
-    public let rateInfo: RateLimit?
+    public let rateInfo: ApiRateLimit?
```
JayNewstrom, line 32: This looks like a breaking API change. Is it possible to do this without the rename? I don't think we'd want to break API compatibility for this change.

mikenachbaur-okta: I think this is a safe enough change, because there's no public initializers for this type, so it's only ever accessed as a property on the APIResponse object.

JayNewstrom: I think we should take a stance here. If we're not adhering to semver, we need to document it. Oktas default is semver.

Sources/AuthFoundation/Network/APIResponse.swift
JayNewstrom: From the requirements document we put together, if this is not available, we delay the min value (I think).

emmanuelogunmuyiwa-okta: Since we don't have a synchronized date support, in the unlikely case where the date is greater than the reset should we delay with the min value?

Tests/AuthFoundationTests/ApiRetryTests.swift
emmanuelogunmuyiwa-okta: Good callout. I see what you're saying. I can use a function n times to represent it.

Sources/AuthFoundation/Network/APIClient.swift
emmanuelogunmuyiwa-okta: I've updated this. Thanks

Sources/AuthFoundation/Network/APIClient.swift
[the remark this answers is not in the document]
emmanuelogunmuyiwa-okta: I've updated this. Thanks

Sources/AuthFoundation/Network/APIResponse.swift
```diff
@@ -29,39 +29,47 @@ public struct APIResponse<T: Decodable>: Decodable {
...
     
     /// The ID for the current request.
     public let requestId: String?
+}
+
+/// Describes information related to the organization's current rate limit.
+public struct ApiRateLimit: Decodable {
```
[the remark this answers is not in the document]
emmanuelogunmuyiwa-okta: Good callout. I've updated this.

Tests/AuthFoundationTests/ApiRetryTests.swift
emmanuelogunmuyiwa-okta: Good callout. I've updated this.

Sources/AuthFoundation/Network/APIResponse.swift
JayNewstrom: Docs stating semver. https://developer.okta.com/code/library-versions/#semantic-versioning

Sources/AuthFoundation/Network/APIResponse.swift
rajdeepnanua-okta: Let's sync up about tradeoffs for including/excluding a minDelay offline. It can be added in a future PR, if that's what we decide later.

Tests/AuthFoundationTests/APIRetryTests.swift
```diff
@@ -0,0 +1,119 @@
...
+            case .failure(let error):
+                XCTAssertEqual(error.errorDescription, APIClientError.statusCode(429).errorDescription)
+            }
+            expect.fulfill()
+        })
+        waitForExpectations(timeout: 2.0) { error in
+            XCTAssertNil(error)
+        }
+    }
+}
```
rajdeepnanua-okta, line 119: Could you also add tests with malformed or missing Date and X-Rate-Limit-Reset headers? Current behavior throws an error in that case, but I think it makes sense to capture that as part of a test.

rajdeepnanua-okta (approved):
Generally looks good. Would be nice to have an additional test with malformed or missing headers, just to capture the expected behavior there. Let's discuss minDelay outside of this PR, and that can be added later if we decide to include it in swift sdk.

Tests/AuthFoundationTests/APIRetryTests.swift
emmanuelogunmuyiwa-okta: I have included test for missing X-Rate-Limit-Reset headers and not missing date. we internally default to Date.nowCoordinated if the date header is missing.

Sources/AuthFoundation/Network/APIResponse.swift
mikenachbaur-okta: I agree, we should discuss this further. I have a few concerns around minDelay, namely that I wouldn't want to introduce artificial delays to a user that might impact their experience.

Sources/AuthFoundation/Network/APIResponse.swift
mikenachbaur-okta: @emmanuelogunmuyiwa-okta can you look into whether or not we can add some sort of typealias or other value with an `@available` depreciation warning to prevent needing to bump the major version?

emmanuelogunmuyiwa-okta: @mikenachbaur-okta I think including the depreciation warning this way might be better that way its easier to adopt.  

```
@available(*, unavailable, renamed: "APIRateLimit")
public struct RateLimit { }
```

mikenachbaur-okta: Yes, but if you use the following, it won't require the user to make any code changes, which ensures that no breaking changes will occur:

```
public struct APIResponse<T: Decodable>: Decodable {
    @available(*, deprecated, renamed: "APIRateLimit")
    public typealias RateLimit = APIRateLimit
```

emmanuelogunmuyiwa-okta: That sounds good. If I use the typealias, I would not add the depreciation warning because Xcode would require a code change with that.

Sources/AuthFoundation/Network/APIClient.swift
```diff
@@ -91,16 +123,25 @@ extension APIClient {
...
+    
+    private func send<T>(_ request: URLRequest,
+                         parsing context: APIParsingContext? = nil,
+                         state: APIRetry.State?,
+                         queue: DispatchQueue = .global(),
+                         completion: @escaping (Result<APIResponse<T>, APIClientError>) -> Void) {
         var urlRequest = request
-        
         willSend(request: &urlRequest)
         session.dataTaskWithRequest(urlRequest) { data, response, httpError in
```
mikenachbaur-okta, line 146: SwiftLint is still complaining about this block. Could you see about either refactoring some of it to be smaller, or we can consider increasing the closure_body_length limit.

Sources/AuthFoundation/Network/APIResponse.swift
```diff
@@ -14,6 +14,8 @@ import Foundation
 
 /// Describes a response from an Okta request, which includes the supplied result, and other associated response metadata.
 public struct APIResponse<T: Decodable>: Decodable {
+    @available(*, deprecated, renamed: "APIRateLimit")
+    public typealias RateLimit = APIRateLimit
     /// Links between response resources.
```
mikenachbaur-okta, line 19: Can you ensure there's an empty line between this?

Sources/AuthFoundation/Network/APIRateLimit.swift
```diff
@@ -0,0 +1,50 @@
...
+    /// The current limit.
+    public let limit: Int
+    
+    /// The rate limit remaining.
+    public let remaining: Int
+
+    /// The time offset from UTC when the rate limit will reset, and a request may be retried.
+    public let reset: TimeInterval
+    
+    public let delay: TimeInterval
```
mikenachbaur-okta, line 26: Can you add some documentation to this property?

Sources/AuthFoundation/Network/APIResponse.swift
emmanuelogunmuyiwa-okta: Good callout. I've updated this.

Sources/AuthFoundation/Network/APIClient.swift
emmanuelogunmuyiwa-okta: Good callout. I've updated this.

Sources/AuthFoundation/Network/APIClient.swift
```diff
@@ -62,24 +62,56 @@ public protocol APIClient {
...
     public func api<T>(client: APIClient, didSend request: URLRequest, received response: APIResponse<T>) {}
+    public func api(client: APIClient, shouldRetry request: URLRequest) -> APIRetry {
+        return .default
+    }
+}
+
+/// List of retry options
+public enum APIRetry {
+    case doNotRetry
+    case retry(maximumRetryCount: Int)
```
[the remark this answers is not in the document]
emmanuelogunmuyiwa-okta: Good callout. I've updated this.

mikenachbaur-okta:
This is looking great. One additional thought, based on our most recent conversation. Since RateLimit information is returned for every request (and not every request needs to be retried), there may not be circumstances where we don't have all the headers for rate limit retry or delay.  For example, if a 429 has _not_ been received, it could still be useful to have the rate limit info associated with the response object.

So we might want to make `delay` optional. We should also ensure, in those circumstances, that delay is at least 1s, to avoid floating-point issues.

emmanuelogunmuyiwa-okta:
@mikenachbaur-okta Thanks for the suggestion, I've made the update as mentioned in your comment here. When a response is validated the rateLimit information is associated with the response object if those headers are present even if it's not a 429 error. 

> This is looking great. One additional thought, based on our most recent conversation. Since RateLimit information is returned for every request (and not every request needs to be retried), there may not be circumstances where we don't have all the headers for rate limit retry or delay. For example, if a 429 has _not_ been received, it could still be useful to have the rate limit info associated with the response object.
> 
> So we might want to make `delay` optional. We should also ensure, in those circumstances, that delay is at least 1s, to avoid floating-point issues.

mikenachbaur-okta (approved):
Looks great! 🎉

PR #99, opened 2022-11-01
Resolve issue where `+` characters would be improperly encoded in query strings

### LINKED ISSUE #98: Character "+" gets replaced with " " (space) when added in "login_hint" as additionalParameter when using signIn from WebAuthentication
### Describe the bug?

I'm trying to sign in using a login hint to open a register view instead of a login view from the IDS the web view). The app has the send a phone number as an additional paramter in the login_hint to do this. The phone number can contain a "+" at the begining.

When adding "login_hint" as additional parameter (either when creating WebAuthentication or in Okta.plist) and using the signIn method of WebAuthentication every "+" character in the "login_hint" will be replaced with a " " (space)

Is the package removing any "+" in the login_hint or is this a problem from the IDS?

Send the string exactly the way it is, with the "+" 

The "+" gets replaced with a " " (space)

### Reproduction Steps?

- Create a WebAuthentication with a login hint
- Use the signIn methob (WebAuthentication.shared.signIn...)
- All the "+" characters in the "login_hint" will be replaced with a " " when the web view is opened

### Additional Information?

A colleague from the android team can send the "+" to the same IDS server

### SDK Version(s)

1.1

### Build Information

_No response_

mikenachbaur-okta:
Resolves #98

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -78,10 +78,12 @@ jobs:
           -sdk iphonesimulator \
           -destination "platform=iOS Simulator,OS=latest,name=iPhone 12" | xcpretty
     - name: Upload Test Results
-      uses: actions/upload-artifact@v2
+      uses: actions/upload-artifact@v3
       with:
         path: ~/TestResults/*.xcresult
         name: XcodeBuild-Results
+        if-no-files-found: warn
+        retention-days: 1
       if: success() || failure()
 
   UITest-WebSignIn:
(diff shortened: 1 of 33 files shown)


mikenachbaur-okta:
@sczitec This should address your issue, if you'd like to review it.

arvindkrishnakumar-okta (approved):

sczitec:
Thanks, that should do it. I will check if it works when the merge is complete and will let you know if there is still a problem

sczitec:
I realized i can download this branch of the package instead of waiting for it to be pushed to master. The "+" is now properly encoded. Thanks

PR #103, opened 2022-11-14
Run tests in parallel to speed up CI performance

mikenachbaur-okta:


rajdeepnanua-okta (approved):

.github/workflows/tests.yaml
```diff
@@ -39,52 +41,77 @@ jobs:
...
-          -destination "platform=iOS Simulator,OS=latest,name=iPhone 12" | xcpretty
+          -destination "$iOS_DESTINATION" | xcpretty
     - name: Upload Test Results
       uses: actions/upload-artifact@v3
       with:
-        path: ~/TestResults/*.xcresult
-        name: XcodeBuild-Results
-        if-no-files-found: warn
+        path: ../iOS-Build.tgz
+        name: iOS-Build
```
rajdeepnanua-okta, line 113: This is no longer uploading test results

.github/workflows/tests.yaml
```diff
@@ -95,40 +122,38 @@ jobs:
...
     - uses: ./.github/actions/setup-secrets
       with:
         okta_plist: "${{ secrets.TEST_OKTA_PLIST }}"
         test_configuration: "${{ secrets.TEST_CONFIGURATION }}"
     - name: Test Web Sign In
       run: |
         xcodebuild \
-            -derivedDataPath ~/Build/DerivedData \
-            -clonedSourcePackagesDirPath ~/Build/ClonedSources \
-            -resultBundlePath ~/TestResults/WebSignIn-iOS.xcresult \
```
rajdeepnanua-okta, line 107: Why was ```-resultBundlePath ~/TestResults/WebSignIn-iOS.xcresult``` removed?

rajdeepnanua-okta (commented):
Generally looks good. I had 2 questions.

.github/workflows/tests.yaml
mikenachbaur-okta: Yes, the test result post-processing was error prone and added several minutes to the build time. I didn't feel like it was adding that much value, so I'm taking it out.

.github/workflows/tests.yaml
mikenachbaur-okta: Exporting the result bundle is only necessary if you want to post-process the test results, which includes screen shots of every step of the UI tests.

PR #112, opened 2023-02-15
Revoking an accessToken should not remove the credential from storage

mikenachbaur-okta:


Samples/AuthenticationFlows.swift
```diff
@@ -38,7 +38,7 @@ func signInWithWebUsingCustomConfiguration() async throws {
                                  redirectUri: redirectUrl)
     
     // Sign in using the above configuration
-    let token = try await auth.start(from: view.window)
+    let token = try await auth.signIn(from: view.window)
```
rajdeepnanua-okta, line 41: How does this change the behavior?

rajdeepnanua-okta (approved):
Looks good. Had a question about the change in sample

Samples/AuthenticationFlows.swift
mikenachbaur-okta: This is just sample code, but a long time ago when I renamed the `start` method to `signIn`, I forgot to update the sample.

rajdeepnanua-okta: Ah got it. Thanks for clarifying!

PR #114, opened 2023-03-09
Resolve issue where the compiler is confused between Swift Concurrency and Block-based functions

mikenachbaur-okta:
The block-based methods currently have optional blocks, based on the idea that delegates would be used to intercept these responses, but to date it appears that this feature is never used. In practice however, this results in the compiler confusing which function should be invoked. For example:

```
let token = try await flow.start()
```

vs
```
flow.start()
```

The first one should use Swift Concurrency, while the second should use the block methods. However, in practice, in recent versions of Swift, the compiler assumes that the first example should use a non-async function, with a default nil completion block.

To address this, the completion block is being changed to nonnull, to make it explicit which function should be invoked by the compiler.


IldarAbdullin-okta (approved):
LGTM!

rajdeepnanua-okta (approved):

PR #117, opened 2023-03-14
Improve Direct Auth error handling

mikenachbaur-okta:


Sources/OktaDirectAuth/DirectAuthFlow.swift
```diff
@@ -22,8 +22,15 @@ public protocol DirectAuthenticationFlowDelegate: AuthenticationDelegate {
...
+    case invalidOTP
+    case oobRejected
+    case invalidChallengeTypesSupported
+    case pollingTimeoutExceeded
     case missingArguments(_ names: [String])
-    case currentStatusMissing
+    case network(error: APIClientError)
+    case oauth2(error: OAuth2Error)
+    case server(error: OAuth2ServerError)
+    case other(error: Error)
```
qianke-okta, line 33: Just tested and I got `.other(error: )` in invalid OTP scenario, and the associated value inside is also a `DirectAuthenticationFlowError. other(error:)`
Screen Shot 2023-03-14 at 2 07 49 PM

Sources/OktaDirectAuth/DirectAuthFlow.swift
```diff
@@ -22,8 +22,15 @@ public protocol DirectAuthenticationFlowDelegate: AuthenticationDelegate {
 }
 
 public enum DirectAuthenticationFlowError: Error {
+    case invalidOTP
+    case oobRejected
+    case invalidChallengeTypesSupported
+    case pollingTimeoutExceeded
     case missingArguments(_ names: [String])
-    case currentStatusMissing
+    case network(error: APIClientError)
```
qianke-okta, line 30: In push rejected scenario I got a `.network(error:)`
Screen Shot 2023-03-14 at 2 08 13 PM

Sources/OktaDirectAuth/DirectAuthFlow.swift
mikenachbaur-okta: Hmm, so that's not working as expected.  OAuth2ServerError should be returned as a FlowError `.server(error:)`. 

Also I think there are server limitations which are causing the wrong error code to be returned. So until those APIs get updated, the client SDK code might require some work-arounds.

PR #118, opened 2023-03-17
Implement automatic time synchronization from server HTTP request responses

mikenachbaur-okta:
This also adds a new property wrapper to implement thread safe updates to primitive values, since concurrent network requests may interact with the `DefaultTimeCoordinator/offset` property simultaneously. I'm not using the SwiftAtomics package here, because I don't want to add an extra interstitial dependency.


Sources/AuthFoundation/Utilities/Property Wrappers/ThreadSafe.swift
```diff
@@ -0,0 +1,46 @@
...
+///
+/// > Important: This is only useful for primitive value types, and will produce unexpected results when used with collection value types, such as dictionaries or arrays.
+@propertyWrapper
+public struct ThreadSafe<T> {
+    private var value: T
+    private let lock = NSLock()
+    private let queue: DispatchQueue
+
+    public var wrappedValue: T {
+        get { queue.sync { value } }
```
IldarAbdullin-okta, line 25: Why we need DispatchQueue here? A bit dangerous since dev can end-up in a deadlock situation(when pass his/her own serial queue and calls getter within queue task context)

Sources/AuthFoundation/Utilities/TimeCoordinator.swift
```diff
@@ -45,24 +45,48 @@ extension Date {
...
     }
     
+    @ThreadSafe
+    private(set) var offset: TimeInterval
+    
+    private var observer: NSObjectProtocol?
+
+    init() {
+        self.offset = 0
+        self.observer = NotificationCenter.default.addObserver(forName: .oauth2ClientCreated,
```
IldarAbdullin-okta, line 60: Don't see `self.observer.removeObserver` counterpart. You still need to call it

IldarAbdullin-okta (commented):
Overall looks good.
Do you wanna to try implement this as an Opt-In feature for devs so they can try it out first? Looks like what you've done is the base implementation now and there is no chance to switch to previous implementation

Sources/AuthFoundation/Utilities/Property Wrappers/ThreadSafe.swift
mikenachbaur-okta: Good point, I suppose I'm being overzealous there.

Sources/AuthFoundation/Utilities/TimeCoordinator.swift
mikenachbaur-okta: D'oh, thanks for catching that. I'm used to using selector-based observers, which automatically remove themselves for you.

mikenachbaur-okta:
The developer has the ability to set / replace the time coordinator by assigning one to `Date.coordinator`. The out-of-the-box default though should do "The Right Thing". To be honest, this is a problem facing almost all mobile apps, and having a solution that just works is the goal.

IldarAbdullin-okta (approved):
LGTM!

shuowu-okta (approved):

jaredperreault-okta (approved):

PR #120, opened 2023-03-22
Improve adding additional parameters during sign in / out

mikenachbaur-okta:
Common workflows within OIDC Redirect authentication for native applications involve different entry-points for “Sign In”, “Register”, skipping “remember me” states, etc. These are controlled through the use of extra parameters to the initial authorize call, which are passed to the sign-in widget. In many cases, the only difference between these authentication modes is the extra parameters (prompt, idp, etc). However, in the current SDK, there is no way to conveniently supply those extra arguments without creating a brand new instance of the authentication client.

This PR adds the ability to include extra arguments to the signIn/signOut WebAuthentication functions, without the need to create an entirely separate client, when all other OIDC/OAuth2 parameters remain the same.

For example:

```swift
let client = try WebAuthentication() // Use the default config
let token = try client.signIn(from: view.window, options: [.prompt(.login)])

// Or
let token = try client.start([with](from: view.window, options): [.idp(url), .custom("key", "value")])
```

Furthermore, there are some instances where the current API accepts values, such as state and nonce at init time, which is counter to the way the API works. These values should be supplied when the flow is started.

### DIFF
--- a/Samples/Shared/Common/ViewControllers/ProfileTableViewController.swift
+++ b/Samples/Shared/Common/ViewControllers/ProfileTableViewController.swift
@@ -169,8 +169,13 @@ class ProfileTableViewController: UITableViewController {
         
         #if !os(tvOS) && canImport(WebAuthenticationUI) && !WEB_AUTH_DISABLED
         if let token = Credential.default?.token {
+            var options: [WebAuthentication.Option]?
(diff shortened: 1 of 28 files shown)


IldarAbdullin-okta (approved):
👍

PR #121, opened 2023-03-23
Introduce the ability to create a Token from just an access_token

mikenachbaur-okta:
There are circumstances where a developer has previously authenticated a user, yet wishes to use the full Mobile SDK. In this situation, the SDK should permit a new Credential to be stored in the client by just supplying the refresh token to the client.

This PR adds a static function to the Token class. From there, it exchanges the refresh token string, using the supplied client, and returns a token or an error.

```swift
class Token {
  // ...
  
  public static func from(refreshToken: String,
                          using client: OAuth2Client, 
                          completion: @escaping(Result<Token, OAuth2Error>) -> Void)
```

An example of this in action can be seen here:

```swift
let client = OAuth2Client(domain: "my-org.okta.com",
                          clientId: "abcd123",
                          scopes: "openid profile offline_access")
let token = try await Token.from(refreshToken: "the_refresh_token",
                                 using: client)
```

### DIFF
--- a/Samples/WebSignIn/WebSignIn (iOS)/WebSignIn/SignInViewController.swift
+++ b/Samples/WebSignIn/WebSignIn (iOS)/WebSignIn/SignInViewController.swift
@@ -15,6 +15,8 @@ import WebAuthenticationUI
 
 class SignInViewController: UIViewController {
     @IBOutlet weak var signInButton: UIButton!
+    @IBOutlet weak var signInWithRefreshButton: UIButton!
+    @IBOutlet weak var ephemeralSwitch: UISwitch!
     @IBOutlet weak var clientIdLabel: UILabel!
 
     let auth = WebAuthentication.shared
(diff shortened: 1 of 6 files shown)


Samples/WebSignIn/WebSignIn (iOS)/WebSignIn/SignInViewController.swift
```diff
@@ -70,4 +74,69 @@ class SignInViewController: UIViewController {
...
+                }
+                return
+            }
+            
+            self.signInWith(refreshToken: string)
+        })
+        present(alert, animated: true)
+    }
+    
+    func signInWith(refreshToken: String) {
```
rajdeepnanua-okta, line 121: Is signInWith visible to the users? Just wondering if users might be confused between signIn and signInWith, but I'm not too familiar with swift syntax here.

rajdeepnanua-okta (approved):

Samples/WebSignIn/WebSignIn (iOS)/WebSignIn/SignInViewController.swift
mikenachbaur-okta: This is just a part of the test app, to streamline the code.

The sample app will auto-detect if a string is copied to the clipboard, and will attempt to trigger a refresh with that.  It makes the sample flow so much smoother to demonstrate. 

Otherwise, it'll prompt the user for the refresh token in an alert dialog.

PR #132, opened 2023-04-21
Prepare OktaDirectAuth for an EA release

mikenachbaur-okta:
This work has been progressing in a feature branch up until now, but the time has come to merge it into `master` and prepare for an upcoming release, since this functionality is now stable.


Sources/AuthFoundation/JWT/Claim.swift
```diff
@@ -280,7 +280,7 @@ public extension HasClaims {
     
     var allClaims: [String] {
         Array([
-            claims.map({ $0.rawValue }),
+            claims.map(\.rawValue),
```
rajdeepnanua-okta, line 283: What does this change do?

Sources/OktaDirectAuth/Internal/Step Handlers/OOBStepHandler.swift
```diff
@@ -0,0 +1,164 @@
...
+    // around differences in the response data.
+    func requestOOBCode(completion: @escaping (Result<OOBResponse, APIClientError>) -> Void) {
+        // Request where OOB is used as the primary factor
+        if let loginHint = loginHint {
+            requestOOBCode(loginHint: loginHint, completion: completion)
+        }
+        
+        // Request where OOB is used as the secondary factor
+        else if let mfaToken = mfaToken {
+            requestOOBCode(mfaToken: mfaToken, completion: completion)
```
rajdeepnanua-okta, line 110: Should the requestOOBCode functions called by this function be private? It seems this function is taking on the responsibility of calling the correct one.

Tests/OktaDirectAuthTests/DirectAuth2FATests.swift
rajdeepnanua-okta, line 1: Should we also have a test for otp here?

Sources/AuthFoundation/JWT/Claim.swift
mikenachbaur-okta: It's using a Swift short-hand to eliminate the need for a closure. Just to tidy up the source code.

Sources/OktaDirectAuth/Internal/Step Handlers/OOBStepHandler.swift
mikenachbaur-okta: I had left them `internal` to simplify testing, but in the end that wasn't necessary. So that's a good idea, marking them `private`.

Tests/OktaDirectAuthTests/DirectAuth2FATests.swift
mikenachbaur-okta: These test files are going to be removed shortly. These were my original integration tests to exercise the API, but they've been superseded by the other unit tests.  These will be repurposed to separate samples.

PR #142, opened 2023-07-06
Expose the underlying server error message in WebAuthentication

### LINKED ISSUE #139: Missing api error type with message that okta-oidc-ios used to provide
### Describe the feature request?

The current SDK is not passing back the error string that is being set during web sign in. Currently we are getting a `authenticationProviderError` error which does not provide any valuable information. In the previous okta-oidc-ios sdk, we used to get an `case api(message: String, underlyingError: Error?)` error which had the appropriate message set by our custom browser signin client. If this is a missing feature, we would like to request it be added. If it's not a missing feature, we would love some help understanding how to get the actual error message provided by the failed web signin.

### New or Affected Resource(s)

WebAuthenticationUI, WebAuthenticationError

### Provide a documentation link

https://okta.github.io/okta-mobile-swift/development/webauthenticationui/documentation/webauthenticationui/webauthenticationerror/authenticationprovidererror(_:)

### Additional Information?

_No response_

mikenachbaur-okta:
This resolves #139

### DIFF
--- a/Sources/OktaOAuth2/Extensions/URL+Extensions.swift
+++ b/Sources/OktaOAuth2/Extensions/URL+Extensions.swift
@@ -1,0 +1,99 @@
+//
+// Copyright (c) 2023-Present, Okta, Inc. and/or its affiliates. All rights reserved.
+// The Okta software accompanied by this notice is provided pursuant to the Apache License, Version 2.0 (the "License.")
+//
+// You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0.
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//
+// See the License for the specific language governing permissions and limitations under the License.
+//
+
+import Foundation
+import AuthFoundation
+
+extension URL {
+    typealias RedirectError = AuthorizationCodeFlow.RedirectError
+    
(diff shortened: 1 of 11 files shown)


shanepadgett:
See reply here: https://github.com/okta/okta-mobile-swift/issues/139#issuecomment-1624488929

This has not fixed the problem yet.

Sources/OktaOAuth2/Extensions/URL+Extensions.swift
```diff
@@ -0,0 +1,99 @@
...
+            }
+        }) else {
+            throw RedirectError.missingQueryArguments
+        }
+        
+        if let state = state {
+            guard query["state"] == state else {
+                throw RedirectError.invalidState(query["state"])
+            }
+        }
```
shanepadgett, line 43: We do not have a query param of state being appended to our redirect url. Because of that, this state check and throw prevents the error description from getting set.

shanepadgett: our redirect looks like com.company.site://auth?error=mappingnotfound&error_description=No%20mappings.

mikenachbaur-okta: @shanepadgett Yes, the state is returned in some circumstances, which happens to be the cases I was testing with. I have an updated commit that removes the `state` value check from the error handling.

shanepadgett: Great. Will test again

shanepadgett: That worked!

PR #145, opened 2023-07-11
Obvious Fix: Change minimum iOS version from 15 to 13 where possible

sebastiangrail:
All async/await APIs are currently only available for iOS 15 but the minimum requirement for async/await is iOS 13. This PR changes the minimum iOS version from 15 to 13 where possible.

### DIFF
--- a/Sources/AuthFoundation/Network/APIRequest.swift
+++ b/Sources/AuthFoundation/Network/APIRequest.swift
@@ -69,7 +69,7 @@ public protocol APIRequest {
     ///   - client: ``APIClient`` the request is being sent to.
     ///   - context: Optional context to use when parsing the response.
     /// - Returns: ``APIResponse`` result of the request.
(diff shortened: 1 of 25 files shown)


mikenachbaur-okta:
Thank you for submitting this. The minimum was set to iOS 15 due to issues with Xcode 13 throwing compiler errors when Swift Concurrency features were used while targeting older iOS versions. These problems seem to have been fixed in Xcode 14, so we can certainly go ahead with this.

Since this SDK supports platforms other than iOS (tvOS, watchOS, and macOS), we'd need to ensure the supported versions match all available platforms though.

@sebastiangrail would you be able to update the supported versions in those @available attributes to support tvOS 13, watchOS 6, and macOS 10.15?

sebastiangrail:
@mikenachbaur-okta Done. Note that the build for tvOS and watchOS is broken for unrelated reasons (tested on Xcode 14.3.1).

PR #146, opened 2023-07-31
Using discovery url

sameh0:
**Enhancements in OAuth2Client & OpenIdConfigurationRequest**

- Properly utilize the passed discovery URL in `OAuth2Client.swift`. Previously, it was only being set but not used.
  
- Eliminated redundant logic: The logic to append `.well-known` to the URL was duplicated in both `OpenIdConfigurationRequest` and the `OAuth2Client.swift` initializer. This has now been streamlined, removing the redundancy from `OpenIdConfigurationRequest`.

---

### DIFF
--- a/Sources/AuthFoundation/OAuth2/OAuth2Client.swift
+++ b/Sources/AuthFoundation/OAuth2/OAuth2Client.swift
@@ -131,7 +131,7 @@ public final class OAuth2Client {
                 
                 openIdConfigurationAction = action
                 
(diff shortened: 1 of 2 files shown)


mikenachbaur-okta:
Thank you for your pull request. This seems like a very straightforward fix to an obvious bug.  I've kicked off the CI process for this change, and if none of the tests fail, I should be able to merge and release it shortly.

PR #152, opened 2023-08-24
Ensure the default URLSession for OAuth2Client is ephemeral

mikenachbaur-okta:
Certain versions of macOS provides a `URLSession.shared` instance that uses a persistent URL cache, whereas newer versions do not.  Since the `/oauth2/v1/keys` endpoint has a long cache expiry (~= 10 days), when keys are rotated on the server, users may not be unable to sign in unless they clear the cache (or delete and reinstall the app).

Since other parts of the SDK explicitly uses an ephemeral session already, this looks an oversight, so it’s good to make our uses of URLSession consistent across the SDK.

### DIFF
--- a/OktaAuthFoundation.podspec
+++ b/OktaAuthFoundation.podspec
@@ -1,7 +1,7 @@
 Pod::Spec.new do |s|
     s.name             = "OktaAuthFoundation"
     s.module_name      = "AuthFoundation"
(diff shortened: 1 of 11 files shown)


IldarAbdullin-okta (approved):
LGTM!

rajdeepnanua-okta (approved):

PR #166, opened 2024-02-06
fix client id matching

colintremblay-okta:
Client IDs for 1st party apps (like Okta Personal's iOS app) have don't match the regex. Since we have access to the client ID, we can first do a check against that instead of the regex, then default the the current behavior if that doesn't work.

This PR:

- covers the failing scenario with a unit test
- fixes the failing scenario
- also fixes a small documentation issue (removes a parameter from the docs that isn't in the actual function)


mikenachbaur-okta (approved):
This looks great! I'll target a feature branch with this PR and will trigger a new PR from there to ensure the UI tests run.

PR #172, opened 2024-02-14
Implement WebAuthn authenticator support in DirectAuth

mikenachbaur-okta:
Adds support for WebAuthn factors within DirectAuth. This also improves upon supplying token parameters, and defining the desired grant type, based on the supplied factor combined with the current status of the user's authentication.

**Note:** the JSONValue class and tests were taken from [okta-idx-swift](https://github.com/okta/okta-idx-swift/blob/master/Sources/OktaIdx/Internal/Implementations/Version1/Extensions/JSONValue.swift). An update to that SDK will made to consume JSONValue from AuthFoundation.


mikenachbaur-okta:
Note: The version of swiftlint updated on Github Actions' runners, so this also includes some changes to address lint warnings.

Sources/AuthFoundation/Utilities/JSONValue.swift
```diff
@@ -0,0 +1,172 @@
...
+
+import Foundation
+
+public enum JSONValueError: Error {
+    case cannotDecode(value: Any)
+}
+
+/// Represent mixed JSON values as instances of `Any`. This is used to expose API response values to Swift native types
+/// where Swift enums are not supported.
+public enum JSONValue: Equatable {
```
mikenachbaur-okta, line 21: **Note:** This class came from the `okta-idx-swift` repo, which will be updated to consume AuthFoundation for this feature in another PR.

Sources/OktaDirectAuth/Internal/Authentication Factors/SecondaryFactor.swift
```diff
@@ -63,6 +84,8 @@ extension DirectAuthenticationFlow.SecondaryFactor: AuthenticationFactor {
             return .otpMFA
         case .oob:
             return .oobMFA
+        case .webAuthn, .webAuthnAssertion(_):
```
santhoshramaraju-okta, line 87: will webauthn in MFA use case be added separately? if so this grant type will probably change to `urn:okta:params:oauth:grant-type:mfa-webauthn` and that will create a problem with resuming the flow using secondary factor similar to the issue https://github.com/okta/okta-mobile-swift/issues/175.

Sources/OktaDirectAuth/WebAuthn/PublicKeyCredentialDescriptor.swift
```diff
@@ -0,0 +1,31 @@
...
+
+extension WebAuthn {
+    /**
+     This dictionary contains the attributes that are specified by a caller when referring to a public key credential as an input parameter to the create() or get() methods. It mirrors the fields of the PublicKeyCredential object returned by the latter methods.
+     
+     - Note: [W3C Reccomendation](https://www.w3.org/TR/webauthn/#dictionary-credential-descriptor)
+     */
+    public struct PublicKeyCredentialDescriptor: Codable {
+        /// This member contains the credential ID of the public key credential the caller is referring to.
+        public let id: Data
```
santhoshramaraju-okta, line 23: Credential ID is sent from server as a string

Sources/OktaDirectAuth/Internal/Authentication Factors/SecondaryFactor.swift
mikenachbaur-okta: No, according to the spec, the grant-type is the same for both primary auth vs MFA auth.  However, you did help me point out one gap, where the list of `challenge_types_supported` will return `urn:okta:params:oauth:grant-type:mfa-webauthn` when initiating a WebAuthn challenge as an MFA factor. So I need to make sure that's supported in this code.

Sources/OktaDirectAuth/Internal/Authentication Factors/SecondaryFactor.swift
```diff
@@ -28,41 +28,73 @@ extension DirectAuthenticationFlow.SecondaryFactor: AuthenticationFactor {
...
+        return result
     }
 
-    var grantType: GrantType {
+    func grantType(currentStatus: DirectAuthenticationFlow.Status?) -> GrantType {
         switch self {
         case .otp:
             return .otpMFA
         case .oob:
             return .oobMFA
```
santhoshramaraju-okta, line 91: can you add the check for mfaToken here too to determine the right grant type to fix the issue with push factor number challenge?

mikenachbaur-okta: I'll do that as a separate PR, since I didn't want to bloat this one any more than it already has.

santhoshramaraju-okta (approved):

Sources/OktaDirectAuth/Internal/Authentication Factors/PrimaryFactor.swift
```diff
@@ -26,61 +26,73 @@ extension DirectAuthenticationFlow.PrimaryFactor {
...
+        case .oob: break
+        case .webAuthn: break
         }
-
+        
+        return result
     }
 
-    var grantType: GrantType {
+    func grantType(currentStatus: DirectAuthenticationFlow.Status?) -> GrantType {
```
rajdeepnanua-okta, line 86: Does this need currentStatus as a function parameter? I don't see that being used in this function

rajdeepnanua-okta (approved):

Sources/OktaDirectAuth/Internal/Authentication Factors/PrimaryFactor.swift
mikenachbaur-okta: The `grantType` property/function is a property of the `AuthenticationFactor` protocol, which is shared by both Primary and Secondary factor types.  The `SecondaryFactor` uses the current status to determine whether or not the grant type should be a primary authentication, or MFA.  

For example, [in this line](https://github.com/okta/okta-mobile-swift/pull/172/files#diff-68b991a9abe502d7dc9f8d2179080fb83e6bd0971e3d84e5ed10968d7fc80a02R93) the code checks to see if the current status is associated with an MFA Token, which changes which grant type should be used.

Sources/AuthFoundation/User Management/Credential.swift
```diff
@@ -229,12 +229,10 @@ public final class Credential: Equatable, OAuth2ClientDelegate {
         let shouldRemove = shouldRemove(for: type)
         
         oauth2.revoke(token, type: type) { result in
-            defer { completion?(result) }
-            
-            guard case .success = result else { return }
-            
             // Remove the credential from storage if the access token was revoked
-            if shouldRemove {
+            if case .success = result,
```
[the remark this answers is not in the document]
mikenachbaur-okta: @baksha97 If the revoke is unsuccessful, the developer could still call `remove()`.  However, if the revoke operation failed due to a network timeout/offline error, the token will still be valid on the server, and removing it will prevent the app from being able to revoke it in the future.

For example, you could do the following:

```
do {
  try await credential.revoke()
} catch {
  // Check to see if the error is recoverable
  if shouldRemoveTheToken {
    try credential.remove()
  }
}
```

Issue #168, opened 2024-02-09
Incorrectly migrates all previous credentials to one client id and tenant. 


baksha97:
### Describe the bug?

Migrating incorrectly captures all previous `StateManager`s and incorrectly registers them to the first registered client id.

### What is expected to happen?

Each credential should be saved with the correct token context/configuration from the original issuing context. 

### What is the actual behavior?

Each credential is saved with a configuration/context of the the first client id registered and migrated. 

### Reproduction Steps?

1. Use `okta-oidc-ios` to log into two+ different okta tenants.
2. Bump to using the new library
3. Register only one client app
4. Call `SDKVersion.migrateIfNeeded`
5. Notice that when retrieving all credentials at a later point in time, all credentials have been assigned the first client app registered. 

### Additional Information?
![image](https://github.com/okta/okta-mobile-swift/assets/15055008/d39537df-8544-409d-9d57-130eae100ea6)

This is a critical bug since the SDK will try to provide a Credential that doesn't belong to the migrated client id and tenant. The SDK will believe tokens are valid for a Client/Tenant and all request will fail due to this mismatch without any remediation from the client possible. 

### SDK Version(s)

All

### Build Information

n/a

PR #176, opened 2024-02-22
Improve Legacy OIDC migration to include the original client ID and issuer

mikenachbaur-okta:
This fix should address #168.

@baksha97 You might want to review this, since it should fix the concerns you pointed out.


rajdeepnanua-okta (approved):

baksha97 (approved):
At a glace, this looks like a great improvement. I don't have the bandwidth to functionally test this new version, so if you can verify this behavior functionally too - that would be great!

Thanks @mikenachbaur-okta

PR #182, opened 2024-03-21
Make OpenIdConfiguration adaptable to a variable number of claims, while making claims themselves generic

mikenachbaur-okta:
Previously the list of OpenIdConfiguration values were hard-coded, and some of the individual claims were marked as non-optional which goes against the public specs.

This update makes the `HasClaims` protocol generic, with the types of claims supported controllable by the conforming type.


Sources/AuthFoundation/Responses/OpenIdProviderMetadata.swift
```diff
@@ -0,0 +1,69 @@
...
+        case uiLocalesSupported
+        case claimsParameterSupported
+        case requestParameterSupported
+        case requestUriParameterSupported
+        case requireRequestUriRegistration
+        case opPolicyUri
+        case opTosUri
+        
+        // Okta-defined additions
+        // https://developer.okta.com/docs/reference/api/oidc/#response-properties-11
```
stevelind-okta, line 57: When we need to support a new custom value, can we still handle that without making a change here?

Tests/AuthFoundationTests/OpenIDConfigurationTests.swift
```diff
@@ -198,6 +198,22 @@ final class OpenIDConfigurationTests: XCTestCase {
...
+        
+        let claimsSupported = try XCTUnwrap(config.claimsSupported)
+        XCTAssertTrue(claimsSupported.contains(.custom("is_private_email")))
+        XCTAssertEqual(config.claimsSupported, [
+            .audience,
+            .email,
+            .emailVerified,
+            .expirationTime,
+            .issuedAt,
+            .custom("is_private_email"),
```
stevelind-okta, line 210: Looks like this is how you do a custom one

stevelind-okta (approved):
Looks good. If I'm reading this right, all the standard claims are now enumerated, including some Okta-specific ones, but there is still an ability to use custom claims. Is that right?

Sources/AuthFoundation/Responses/OpenIdProviderMetadata.swift
mikenachbaur-okta: Correct, the `.custom(_)` enum case is used for undefined or unmatched cases. If we want to assign others for convenience in the future we can.

mikenachbaur-okta:
Correct @stevelind-okta, the generic `HasClaims` protocol is used in a variety of places since they're fairly common within OIDC. This update makes them more generally applicable, and adapts the OpenIdConfiguration struct to be more adaptive, and adhere to the spec closer. This should go a long way to supporting alternative authorization servers.

Sources/AuthFoundation/JWT/Enums/JWTClaim.swift
```diff
@@ -229,62 +232,24 @@ public enum Claim: Codable {
...
     
-    /// All claims, across both standard ``claims`` and ``customClaims``.
-    var allClaims: [String] { get }
+    /// Indicates the liklihood of whether or not this appears to be a real user. Used predominantly by Sign In With Apple.
+    case realUserStatus
     
-    /// Raw paylaod of claims, as a dictionary representation.
-    var payload: [String: Any] { get }
+    /// Indicates if the email address provided is a proxied address. Used predominantly by Sign In With Apple.
+    case isPrivateEmail
```
stevelind-okta, line 243: `isPrivate` vs `realUser` if we aren't using `is` for boolean should we just do `privateEmail`?

Sources/AuthFoundation/JWT/Enums/JWTClaim.swift
```diff
@@ -229,62 +232,24 @@ public enum Claim: Codable {
...
-    /// Returns the collection of claims this object contains.
-    ///
-    /// > Note: This will only return the list of official claims defined in the ``Claim`` enum. For custom claims, please see the ``customClaims`` property.
-    var claims: [Claim] { get }
     
-    /// Returns the collection of custom claims this object contains.
-    ///
-    /// Unlike the ``claims`` property, this returns values as strings.
-    var customClaims: [String] { get }
+    /// Indicates whether the transaction is on a nonce-supported platform. If you sent a nonce in the authorization request but do not see the nonce claim in the ID token, check this claim to determine how to proceed. Used predominantly by Sign In With Apple.
```
stevelind-okta, line 236: `platform` or `protocol`?

stevelind-okta (approved):
👍

Sources/AuthFoundation/JWT/Enums/JWTClaim.swift
mikenachbaur-okta: This is referring to the Authorization Server platform (this is mostly copy/pasted from the OIDC spec itself).

Sources/AuthFoundation/JWT/Enums/JWTClaim.swift
mikenachbaur-okta: These are taken from the Apple OIDC extensions, and are the camel-case versions of the claims themselves, so I didn't want to tweak them too much.

jaredperreault-okta (approved):

PR #190, opened 2024-05-02
EdDSA in openid-configuration support

sdkdimon:
Added EdDSA case to support openid-configuration where EdDSA presented in "id_token_signing_alg_values_supported" field.


mikenachbaur-okta:
Hi @sdkdimon! Thank you for this PR. I figured it would be more flexible to address this with support for "other" algorithms to prevent future problems, so I've created #192, if you'd like to review it.

PR #193, opened 2024-05-17
Prevent the app from being suspended while refreshing a token

mikenachbaur-okta:


colintremblay-okta (approved):
looks great! iirc, this functionality isn't needed/isn't possible on mac because the concept of backgrounding doesn't really apply

mikenachbaur-okta:
Correct @colintremblay-okta, those APIs are only available on the listed platforms.  I left the class in place for other unsupported platforms as a stub though, just in case we need to add support for equivalent capabilities on macOS, Linux, etc.

PR #197, opened 2024-07-10
Refresh token expiration field

sdkdimon:
- added refreshExpiresIn field if present in response

### DIFF
--- a/Sources/AuthFoundation/Migration/Migrators/OIDCLegacyMigrator.swift
+++ b/Sources/AuthFoundation/Migration/Migrators/OIDCLegacyMigrator.swift
@@ -186,6 +186,7 @@ extension SDKVersion.Migration {
                               accessToken: accessToken,
                               scope: scope,
                               refreshToken: tokenResponse.refreshToken,
+                              refreshExpiresIn: tokenResponse.refreshTokenExpirationDate?.timeIntervalSinceNow,
                               idToken: idToken,
                               deviceSecret: nil,
                               context: Token.Context(configuration: configuration,
@@ -252,6 +253,7 @@ extension SDKVersion.Migration {
                 @objc let tokenType: String?
                 @objc let idToken: String?
                 @objc let refreshToken: String?
+                @objc let refreshTokenExpirationDate: Date?
                 @objc let scope: String?
                 @objc let additionalParameters: [String: String]?
 
@@ -261,6 +263,7 @@ extension SDKVersion.Migration {
                     accessToken = coder.decodeObject(forKey: "access_token") as? String
                     accessTokenExpirationDate = coder.decodeObject(forKey: "expires_in") as? Date
                     refreshToken = coder.decodeObject(forKey: "refresh_token") as? String
+                    refreshTokenExpirationDate = coder.decodeObject(forKey: "refresh_expires_in") as? Date
                     tokenType = coder.decodeObject(forKey: "token_type") as? String
                     idToken = coder.decodeObject(forKey: "id_token") as? String
                     scope = coder.decodeObject(forKey: "scope") as? String
--- a/Sources/AuthFoundation/Token Management/Internal/Token+Internal.swift
+++ b/Sources/AuthFoundation/Token Management/Internal/Token+Internal.swift
@@ -32,6 +32,7 @@ extension Token {
                      accessToken: accessToken,
                      scope: scope,
                      refreshToken: refreshToken ?? token.refreshToken,
+                     refreshExpiresIn: refreshExpiresIn ?? token.refreshExpiresIn,
                      idToken: idToken,
                      deviceSecret: deviceSecret ?? token.deviceSecret,
                      context: context)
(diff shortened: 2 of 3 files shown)


mikenachbaur-okta:
@sdkdimon Interesting, I've never seen the `refresh_expires_in` Token Response parameter in any of the OAuth2 RFCs. Do you know which Authorization Servers are returning this value? The specs only reference the `expires_in` property which references the TTL for the access token.

Typically other tokens' expiration can be returned by introspecting the token, e.g. using [Credential.introspect](https://okta.github.io/okta-mobile-swift/development/authfoundation/documentation/authfoundation/credential/introspect(_:completion:)).

sdkdimon:
I am receiving such response from Keycloak:

```
{
	"access_token": "deadbeef.beef",
	"expires_in": 1800,
	"refresh_expires_in": 3600,
	"refresh_token": "deadbeef.beef",
	"token_type": "Bearer",
	"id_token": "deadbeef.beef",
	"not-before-policy": 0,
	"session_state": "deadbeef.beef",
	"scope": "openid email profile"
}
```
Where  ```"refresh_expires_in"``` field is presented. But in my case i can just parse ```"refresh_token"``` field with JWT struct 
```
let refreshJWT = try? JWT("deadbeef.beef")
```
 and get expiration date from there. Also JWT parse based solution is better because in Keyckloak response ```"issued_at"``` parameter is not presented but it is presented in the token itself.
So i am not sure if this pull request is needed, i made this pull request just before i discovered JWT parse solution.

mikenachbaur-okta:
Ah I see.  Are you using Keycloak in combination with Okta? Or are you using this SDK stand-alone? (Both options are awesome in my mind, I'm just curious about the context).

The `refresh_token` and `access_token` values are defined by the OAuth2 specs to be opaque strings, and while many IAM implementations use JWTs for these, it's not a guarantee that one can rely upon.

Perhaps the best approach is to make the Token object support the [HasClaims protocol](https://okta.github.io/okta-mobile-swift/development/authfoundation/documentation/authfoundation/hasclaims), which would allow for custom properties to be pulled out of the raw response.  This way the SDK can validate all the required security & validation steps defined in the OAuth2/OIDC specs, and any non-standard values can still be extracted.  This pattern is being applied to [JWT](https://okta.github.io/okta-mobile-swift/development/authfoundation/documentation/authfoundation/jwt/), [OpenIdConfiguration](https://okta.github.io/okta-mobile-swift/development/authfoundation/documentation/authfoundation/openidconfiguration/), and other similar types.  So it stands to reason that Token can use this as well.

I wrote some detailed documentation ([Working with Claims](https://okta.github.io/okta-mobile-swift/development/authfoundation/documentation/authfoundation/workingwithclaims)) on how claims are used within the SDK, particularly since they're sprinkled throughout the OAuth2 public specs, so having a generic and common pattern for using these across the SDK is imperative.

For example, I recently changed [the OpenIdConfiguration struct](https://github.com/okta/okta-mobile-swift/blob/master/Sources/AuthFoundation/Responses/OpenIdConfiguration.swift) to support [JSONClaimContainer](https://okta.github.io/okta-mobile-swift/development/authfoundation/documentation/authfoundation/jsonclaimcontainer) which defines an object that can contain claims.  Its ClaimType generic is defined as being [OpenIdProviderMetadata](https://github.com/okta/okta-mobile-swift/blob/master/Sources/AuthFoundation/Responses/OpenIdProviderMetadata.swift), which includes some pre-defined enum values for common keys defined by the spec, but custom keys can be retrieved using the [allClaims](https://okta.github.io/okta-mobile-swift/development/authfoundation/documentation/authfoundation/hasclaims/allclaims) or [customClaims](https://okta.github.io/okta-mobile-swift/development/authfoundation/documentation/authfoundation/hasclaims/customclaims-47hcx) properties.

If that's something you'd be interested in, it should be pretty straightforward to add to your PR.

mikenachbaur-okta:
I've submitted #204 to adapt the Token object to support custom and non-standard properties provided in the Token response, by leveraging the `HasClaims` protocol within the Token itself. This way, the original JSON response is preserved within the token, allowing custom claims (such as `refresh_expires_in`) to be accessed. @sdkdimon This should provide a more flexible way to handle your use-case, as well as supporting other custom responses.

For example, you can do the following to request the value you're looking for:

```
guard let credential = Credential.default else { return } // Or however else you get your credential

if let expiresIn: TimeInterval = credential.token["refresh_expires_in"],
  let expiresDate = credential.token.issuedAt.addingTimeInterval(expiresIn)
{
  // Use the expiresDate
}
```

Alternatively, you can add an extension to Token to do the same thing:

```
import AuthFoundation

extension Token {
  public var refreshExpiresIn: TimeInterval? { self["refresh_expires_in"] }
  public var refreshExpiresAt: Date? {
    guard let expiresIn = refreshExpiresIn else { return nil }
    return issuedAt.addingTimeInterval(expiresIn)
  }
}
```

sdkdimon:
@mikenachbaur-okta great thanks.  I am using okta-mobile-swift just with Keycloak withou Okta, answering on a previous post question.

mikenachbaur-okta:
@sdkdimon I'm going to close this in favor of #204, but please let me know if this doesn't meet your needs.

Issue #198, opened 2024-07-12
Inconsistent Access Permissions When Restoring from Backup


baksha97:
### Describe the bug?

I've encountered an edge case where restoring from a backup causes inconsistent access permissions due to differing configurations between token storage and metadata storage in `KeychainTokenStorage.swift`.

The access settings and security settings for tokens and metadata are configured differently, leading to inconsistencies:

1. **Accessibility**:
   - **Token Storage (`item`)**: Uses variable `accessibility` (Line 87).
   - **Metadata Storage (`metadataItem`)**: Fixed `.afterFirstUnlock` (Line 98).

2. **Access Control**:
   - **Token Storage (`item`)**: Includes `accessControl` (Line 89).
   - **Metadata Storage (`metadataItem`)**: No `accessControl` (Line 96).

Additionally, the built-in defaults for the SDK itself are:
```swift
public static var standard: [Security] = [.accessibility(.afterFirstUnlockThisDeviceOnly)]
```
This indicates that the default behavior with the fixed accessibility (`.afterFirstUnlock`) for metadata leads to a default mismatching. 

When querying for all credentials, the list contains an ID that shouldn't be there. This inconsistency causes all querying of logins to fail permanently on a user's device (notice how this `try` error propagation short-circuits the `compactMap` - did we mean to use `try?` instead?). This likely results from the inconsistency between the token and metadata access settings, causing only the metadata to be available during backup restoration.

```swift
 func find(where expression: @escaping (Token.Metadata) -> Bool,
              prompt: String? = nil,
              authenticationContext: TokenAuthenticationContext? = nil) throws -> [Credential]
    {
        try allIDs
            .map(tokenStorage.metadata(for:))
            .filter(expression)
            .compactMap({ metadata in
                try self.with(id: metadata.id, prompt: prompt, authenticationContext: authenticationContext)
            })
    }
```

### What is expected to happen?

Both token storage and metadata storage should have consistent access permissions and security settings, ensuring that all credentials are accessible and querying for logins does not fail.

### What is the actual behavior?

Tokens and metadata have inconsistent access permissions, leading to access issues and IDs appearing in the list when they shouldn't. This causes **all** usage of `Credential.find` to fail permanently on a user's device.

### Reproduction Steps

1. Store a token and its metadata using `KeychainTokenStorage`.
2. Create a backup of the keychain.
3. Restore the backup on a **new device.**
4. Query for all credentials.

[...]

PR #199, opened 2024-07-12
Allows Keychain read failures to be ignored

baksha97:
Alleviates invalid state for the user created by bug #198.

This allows failing queries into storage (Keychain) to be ignored, for the purpose of allowing `Credential.find` calls to provide as many valid `Credential`s as it can without it being short-circuited on a single failure. 

If the metadata or the item itself is not found in storage, we can disregard the entire `Credential` instead of throwing an error for the entire query. This allows applications to query for a `Credential`, even if the persistence layer has an invalid state for any given `Credential`.

### DIFF
--- a/Sources/AuthFoundation/User Management/Internal/CredentialCoordinatorImpl.swift
+++ b/Sources/AuthFoundation/User Management/Internal/CredentialCoordinatorImpl.swift
@@ -80,12 +80,10 @@ final class CredentialCoordinatorImpl: CredentialCoordinator {
               prompt: String? = nil,
               authenticationContext: TokenAuthenticationContext? = nil) throws -> [Credential]
     {
-        try allIDs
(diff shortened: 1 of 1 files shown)


mikenachbaur-okta:
Thank you for the submission! After considering this, and trying a few options, I felt that using `compactMap` in combination with `try?` could obscure legitimate errors.  Instead, I created PR #200 which solves this problem by:

1. Ensuring the metadata and item keychain entries properly match their synchronization status, and
2. Only returning token IDs from `allIDs` if both their metadata and item entries are available.

mikenachbaur-okta:
Thank you for the submission, though since this is resolved in #200, I'm going to close this PR. This was very helpful in narrowing down the root cause of the problem though, so please keep the suggestions coming!

PR #200, opened 2024-07-17
Ensure token item and metadata keychain accessibility settings are consistent.

mikenachbaur-okta:
This addresses #198 by:

1. Ensuring the token metadata is stored with the appropriate accessibility settings, respecting the parent item's iCloud sync setting.
2. Returning token IDs only if both their item and metadata keychain entries can be found.
3. Establishing the `Credential.Security.isDefaultSynchronizable` property, allowing the default identifier to be shared.

This also adds unit test coverage to validate these changes.

### DIFF
--- a/Sources/AuthFoundation/Token Management/Internal/KeychainTokenStorage.swift
+++ b/Sources/AuthFoundation/Token Management/Internal/KeychainTokenStorage.swift
@@ -99,7 +99,7 @@ final class KeychainTokenStorage: TokenStorage {
                                  value: data)
         
         let metadataAccessibility: Keychain.Accessibility
-        if (accessibility.isSynchronizable) {
+        if accessibility.isSynchronizable {
             metadataAccessibility = .afterFirstUnlock
         } else {
             metadataAccessibility = .afterFirstUnlockThisDeviceOnly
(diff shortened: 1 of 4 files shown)


rajdeepnanua-okta (approved):

Sources/AuthFoundation/Token Management/Internal/KeychainTokenStorage.swift
```diff
@@ -46,10 +46,16 @@ final class KeychainTokenStorage: TokenStorage {
...
+            let itemIDs = try Keychain
+                .Search(service: KeychainTokenStorage.serviceName)
+                .list()
+                .sorted(by: { $0.creationDate < $1.creationDate })
+                .map(\.account)
+            let metadataIDs = try Keychain
                 .Search(service: KeychainTokenStorage.metadataName)
                 .list()
                 .map(\.account)
+            return itemIDs.filter { metadataIDs.contains($0) }
```
baksha97, line 58: Does this prevent the `find` function from short circuiting for users already missing items/tokens?

mikenachbaur-okta: @baksha97 correct, this ensures that only tokens that have both their metadata and item keychain entries available to the device will be returned.  Therefore, the `find` and other functions will succeed.

This will still allow legitimate errors (such as an unsuccessful biometric prompt) to flow through.

PR #204, opened 2024-08-02
Extend Token with `HasClaims` for custom claims

mikenachbaur-okta:
The IETF RFCs for Tokens not only include a couple extra properties that were not previously addressed, but it states that the responses may be extended with custom information. Authorization servers, such as Keycloak, includes metadata in the token response indicating refresh token expiration, or other data.

Since Token responses can be extended, it stands to reason that these objects should conform to `HasClaims` just like other models, to allow for them to be adapted to different authentication scenarios.

This update further extends the support for Claims to resolve edge-cases in how convertible values were handled, unifying support around mapping claims to scalar values, arrays, or dictionaries, and cleaning up how JSON objects are mapped.

Finally, since the result of refresh operations often doesn't include data provided in the initial token exchange request (such as `device_secret`), the process for merging tokens during refresh operations has been moved to a protocol. This is not yet public, but this may be exposed in the future.

This PR:
1. Alters Token to conform to HasClaims so its raw JSON token response data is stored and persisted
2. Defines properties defined by the OAuth2 specs, including required values (such as Token Type, Access Token, etc)
3. Includes version migration support to handle loading and parsing of tokens stored with previous versions of the SDK
4. Refines and makes the HasClaims value functions work consistently (since there were gaps before)
5. Cleans up JSON handling, as well as documentation updates


Sources/AuthFoundation/JWT/Extensions/ClaimConvertable+Extensions.swift
```diff
@@ -12,25 +12,42 @@
...
+@_documentation(visibility: private)
 extension Int: ClaimConvertable {}
+
+@_documentation(visibility: private)
 extension Double: ClaimConvertable {}
+
+@_documentation(visibility: private)
 extension Float: ClaimConvertable {}
-extension Array<String>: ClaimConvertable {}
-extension Dictionary<String, String>: ClaimConvertable {}
```
stevelind-okta, line 21: Does JWTClaim take care of array and dictionary?

Sources/AuthFoundation/JWT/Extensions/ClaimConvertable+Extensions.swift
```diff
@@ -12,25 +12,42 @@
 
 import Foundation
 
+@_documentation(visibility: private)
```
stevelind-okta, line 15: curious, so this hides the extension from generated docs or from Xcode as well?

Tests/AuthFoundationTests/CredentialRevokeTests.swift
```diff
@@ -19,19 +19,19 @@ final class CredentialTests: XCTestCase {
...
-                      accessToken: "abcd123",
-                      scope: "openid",
-                      refreshToken: "refresh123",
-                      idToken: nil,
-                      deviceSecret: "device123",
-                      context: Token.Context(configuration: .init(baseURL: URL(string: "https://example.com/oauth2/default")!,
-                                                                  clientId: "clientid",
-                                                                  scopes: "openid"),
-                                             clientSettings: [ "client_id": "foo" ]))
+    let token = try! Token(id: "TokenId",
```
stevelind-okta, line 22:
```suggestion
    guard let token = try? Token(id: "TokenId", ... else { XCTFail() }?
```

Tests/TestCommon/MockToken.swift
```diff
@@ -68,19 +68,19 @@ extension Token {
...
-                     accessToken: JWT.mockAccessToken,
-                     scope: scopes,
-                     refreshToken: refreshToken,
-                     idToken: idToken,
-                     deviceSecret: deviceSecret,
-                     context: Token.Context(configuration: .init(baseURL: URL(string: "https://example.com/oauth2/default")!,
-                                                                 clientId: "clientid",
-                                                                 scopes: scopes),
-                                            clientSettings: [ "client_id": "clientid" ]))
+        return try! Token(id: "TokenId",
```
stevelind-okta, line 71: Do we want to test anything about these tokens once created? Or is the constructor enough of a test by itself?

stevelind-okta (approved):
Cool refactor, nice docs as usual!

I am having trouble following what happened to nested Array and Dictionary types. Is there some way they are being recursively handled as ClaimConvertible types?

Sources/AuthFoundation/JWT/Extensions/ClaimConvertable+Extensions.swift
mikenachbaur-okta: From the generated docs. Many of these just clutter up the docs. For reference, here's [the list of underscored attributes](https://github.com/swiftlang/swift/blob/main/docs/ReferenceGuides/UnderscoredAttributes.md).

Tests/AuthFoundationTests/CredentialRevokeTests.swift
mikenachbaur-okta: That's in a class property, not a function block, so I opted to simply force-unwrapping since it's in a unit test.  Though I could move initializing the Token to the `setUpWithError` function instead.

Tests/TestCommon/MockToken.swift
mikenachbaur-okta: There's the `Tests/AuthFoundationTests/TokenTests.swift` class that exercises the various ways the token can be initialized, including migrating from V1 data.

This mock token utility is simply there to return tokens for use in other tests. 

Force unwrapping makes me feel icky, but as a utility class, if I required all the other tests using this to `try`, it would make the tests messier.

Sources/AuthFoundation/JWT/Extensions/ClaimConvertable+Extensions.swift
mikenachbaur-okta: It's unnecessary since the accessors for working with claims (e.g. `value(...)` or `subscript` methods) address arrays or dictionaries already.

The goal for ClaimConvertable is to map some JSON primitive supplied from the server into a more developer-friendly value (e.g. from a string to `URL`, or to an enum, etc.). As a result, arrays and dictionaries themselves aren't all that interesting to be converted.

jaredperreault-okta (approved):

PR #206, opened 2024-08-07
Implement RFC7523 with JWTAuthorizationFlow

mikenachbaur-okta:
Adds support for the JWT Bearer Authorization flow. This enables clients to authenticate using a JWT signed from a trusted key as an assertion.

### DIFF
--- a/Sources/OktaOAuth2/Authentication/JWTAuthorizationFlow.swift
+++ b/Sources/OktaOAuth2/Authentication/JWTAuthorizationFlow.swift
@@ -1,5 +1,5 @@
 //
-// Copyright (c) 2021-Present, Okta, Inc. and/or its affiliates. All rights reserved.
+// Copyright (c) 2024-Present, Okta, Inc. and/or its affiliates. All rights reserved.
 // The Okta software accompanied by this notice is provided pursuant to the Apache License, Version 2.0 (the "License.")
 //
 // You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0.
@@ -13,17 +13,13 @@
 import Foundation
 import AuthFoundation
 
-/// An authentication flow class that implements the Resource Owner Flow exchange.
-///
-/// This simple authentication flow permits a suer to authenticate using a simple username and password. As such, the configuration is straightforward.
-///
-/// > Important: Resource Owner authentication does not support MFA or other more secure authentication models, and is not recommended for production applications. Please use the DirectAuth SDK's DirectAuthenticationFlow class instead.
-public class ResourceOwnerFlow: AuthenticationFlow {
+/// An authentication flow class that implements the JWT Authorization Bearer Flow, for authenticating users using JWTs signed by a trusted key.
+public class JWTAuthorizationFlow: AuthenticationFlow {
     /// The OAuth2Client this authentication flow will use.
     public let client: OAuth2Client
     
     /// Indicates whether or not this flow is currently in the process of authenticating a user.
-    /// ``ResourceOwnerFlow/init(issuer:clientId:scopes:)``
+    /// ``JWTAuthorizationFlow/init(issuer:clientId:scopes:)``
     public private(set) var isAuthenticating: Bool = false {
         didSet {
             guard oldValue != isAuthenticating else {
@@ -52,6 +48,8 @@ public class ResourceOwnerFlow: AuthenticationFlow {
                                        scopes: scopes))
     }
     
+    /// Initializer to construct an authentication flow from an OAuth2Client.
+    /// - Parameter client: `OAuth2Client` instance to authenticate with.
     public init(client: OAuth2Client) {
         // Ensure this SDK's static version is included in the user agent.
         SDKVersion.register(sdk: Version)
@@ -78,12 +76,11 @@ public class ResourceOwnerFlow: AuthenticationFlow {
                   scopes: config.scopes)
     }
     
-    /// Authenticates using the supplied username and password.
+    /// Authenticates using the supplied JWT bearer assertion.
     /// - Parameters:
-    ///   - username: Username
-    ///   - password: Password
+    ///   - assertion: JWT Assertion
     ///   - completion: Completion invoked when a response is received.
-    public func start(username: String, password: String, completion: @escaping (Result<Token, OAuth2Error>) -> Void) {
+    public func start(with assertion: JWT, completion: @escaping (Result<Token, OAuth2Error>) -> Void) {
         isAuthenticating = true
 
         client.openIdConfiguration { result in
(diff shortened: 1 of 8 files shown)


rajdeepnanua-okta (approved):

Sources/OktaOAuth2/Authentication/JWTAuthorizationFlow.swift
```diff
@@ -0,0 +1,151 @@
...
+        didSet {
+            guard oldValue != isAuthenticating else {
+                return
+            }
+            
+            if isAuthenticating {
+                delegateCollection.invoke { $0.authenticationStarted(flow: self) }
+            } else {
+                delegateCollection.invoke { $0.authenticationFinished(flow: self) }
+            }
```
stevelind-okta, line 33: Do you want this to call inline or should it be dispatched? Maybe consider changing to a function instead of a side effect of a variable change?

Tests/OktaOAuth2Tests/JWTAuthorizationFlowTests.swift
```diff
@@ -0,0 +1,165 @@
...
+    typealias Flow = JWTAuthorizationFlow
+    
+    var token: Token?
+    var error: OAuth2Error?
+    var url: URL?
+    var started = false
+    var finished = false
+    
+    func authenticationStarted<Flow>(flow: Flow) {
+        started = true
```
stevelind-okta, line 28: should we record the flow?

Tests/OktaOAuth2Tests/JWTAuthorizationFlowTests.swift
```diff
@@ -0,0 +1,165 @@
...
+        
+        let request = try XCTUnwrap(urlSession.requests.first(where: { request in
+            request.url?.lastPathComponent == "token"
+        }))
+
+        XCTAssertEqual(request.url?.absoluteString, "https://example.okta.com/oauth2/v1/token")
+        XCTAssertEqual(request.bodyString, "assertion=\(JWT.mockIDToken)&client_id=clientId&grant_type=urn:ietf:params:oauth:grant-type:jwt-bearer&scope=profile+openid")
+    }
+    
+#if swift(>=5.5.1) && !os(Linux)
```
stevelind-okta, line 146: do we ci infra to test 'old swift' / iOS 12? Just curious

Tests/OktaOAuth2Tests/TokenExchangeFlowTests.swift
```diff
@@ -122,6 +131,15 @@ final class TokenExchangeFlowTests: XCTestCase {
...
+        
+        let request = try XCTUnwrap(urlSession.requests.first(where: { request in
+            request.url?.lastPathComponent == "token"
+        }))
+
+        XCTAssertEqual(request.url?.absoluteString, "https://example.okta.com/oauth2/v1/token")
+        XCTAssertEqual(request.bodyString, "actor_token=secret&actor_token_type=urn:x-oath:params:oauth:token-type:device-secret&audience=api:%2F%2Fdefault&client_id=clientId&grant_type=urn:ietf:params:oauth:grant-type:token-exchange&scope=profile+openid+device_sso&subject_token=id_token&subject_token_type=urn:ietf:params:oauth:token-type:id_token")
     }
     
 #if swift(>=5.5.1) && !os(Linux)
```
stevelind-okta, line 145: do we run tests on linux sometimes?

stevelind-okta (approved):
Love your attention to documentation! Testament to a good design how simply the new flow fits into the SDK.

Some minor comments, no blockers

One question, do we think it's worthwhile to use the new Swift APIs if we have to ifdef them or should we consider dropping support below swift 5? Generally apple doesn't allow folks to submit with old Xcode or old Swift IIRC. The ifdefs make the otherwise very readable code more complex

Sources/OktaOAuth2/Authentication/JWTAuthorizationFlow.swift
mikenachbaur-okta: The SDK affords developers the option to interact with the APIs using either completion blocks or async results, as well as responding to delegate calls.  So I've tried to centralize the notification logic as much as possible to ensure consistency.

I agree with your point though. I have some in-flight work that I haven't completed, but it introduces support for dispatch queues to be used in delegate calls, as well as propagates task cancellation handles up through the SDK, to allow async operations to be cancelled. That involves quite a bit of concurrency testing, so while the work is done, I wasn't happy with the level of test coverage, so I've been holding off on it.

Since I'll be pushing out a major release sometime soon, I might take the opportunity to finish that work.

Tests/OktaOAuth2Tests/JWTAuthorizationFlowTests.swift
mikenachbaur-okta: This DelegateRecorder class is used in testing, to observe invocations of the authentication delegate, and to validate that the appropriate calls are made, and in the correct order.

I really need to make this class support generics so it can be reused elsewhere (e.g. there's almost identical copies of this for Resource Owner Flow, Token Exchange Flow, etc), but I didn't feel that was a priority at this time.

Tests/OktaOAuth2Tests/JWTAuthorizationFlowTests.swift
mikenachbaur-okta: No we don't. [Our support policy](okta-mobile-swift-task-cancellation) says that we support older versions, but it's on a best-effort basis.  e.g. we won't actively deprecate old versions if we don't have to, but if we find it necessary to do so (e.g. a new feature would be cumbersome to backport), then changing the minimum-supported OS version is considered a non-breaking change.

When this SDK was first written, even though Swift Concurrency was supported, there were bugs that prevented Xcode from compiling if the async operator were used. The Linux Swift runtime at the time also didn't support Swift Concurrency, so that's the reason for these `#if` checks. 

I really should pull them all out, and update the minimum Swift version (which is needed to add support for visionOS), but again, I'll wait for a major version change to do that.

Tests/OktaOAuth2Tests/TokenExchangeFlowTests.swift
mikenachbaur-okta: We used to have tests running on Linux, but Github Actions' runner images had a problem with that, and I haven't had the impetus to check on that again.  Linux support is currently considered experimental.

Thanks for the reminder though, if I get a few spare cycles, I'll try to test this out.

mikenachbaur-okta:
Thanks for the comments @stevelind-okta! Yeah, the SDK requires Swift 5+, but Swift Concurrency was only introduced in Swift 5.5.1, which is why we have both GCD and Swift Concurrency versions of its functions.

We might do something similar for the Kotlin SDK, introducing a Java-compat layer, since the SDK makes heavy use of Kotlin Coroutines which aren't available in Java.

PR #207, opened 2024-08-15
Update minimum Swift, Xcode, and platform versions

mikenachbaur-okta:
As per [this SDK's support policy](https://github.com/okta/okta-mobile-swift?tab=readme-ov-file#support-policy), this updates the minimum supported version of Xcode to 15, and as a result updates the supported version of Swift to 5.9, and the minimum supported platform versions.

Furthermore, this update reintroduces experimental support on Linux.


mikenachbaur-okta:
Unfortunately git didn't detect the move from `UnsafeLock.swift` to `Lock.swift`, so it's showing up as a deletion/add. 🤷🏼‍♀️

Sources/AuthFoundation/Utilities/JSONValue.swift
```diff
@@ -251,13 +251,13 @@ extension JSON: Codable {
 
 fileprivate extension NSNumber {
     var isFloatingPoint: Bool {
-        let type = CFNumberGetType(self as CFNumber)
-        switch type {
-        case .floatType, .float32Type, .float64Type, .cgFloatType, .doubleType:
+        if strcmp(objCType, "f") == 0 ||
+            strcmp(objCType, "d") == 0
+        {
```
stevelind-okta, line 256: interesting, how sure are you that this will work permanently?

stevelind-okta (approved):
What does our testing look like for linux and visionOS?

Sources/AuthFoundation/Utilities/JSONValue.swift
mikenachbaur-okta: https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ObjCRuntimeGuide/Articles/ocrtTypeEncodings.html

This has been around within Objective-C since the dawn of time, so it's unlikely to change.  I would use the `@encode` feature, but that's not available in Swift, so it was more straightforward to use the string equivalents.

mikenachbaur-okta:
@stevelind-okta Linux testing is currently limited to running `swift test` on ubuntu-latest, but no E2E automated test flows (yet).  visionOS is being tested with Xcode automated tests against visionOS 1.0, but no E2E flows right now either.

To date we haven't really had any matrix / legacy version testing as part of the CI process, so this PR is the first to introduce parallel unit tests against different OS platforms and versions. I'm considering doing the same for our E2E UI tests, but I'd capture that as a separate story.

rajdeepnanua-okta (approved):

PR #211, opened 2024-10-26
Massive project reorganization and thread-safety updates for Swift 6.

mikenachbaur-okta:
* Break up AuthFoundation into smaller convenience libraries
* Include macros and types for implementing thread safety and consistency
* Shore up credential storage & lifecycle to prevent existing race conditions around `Credential.default`


chasemac:
How soon will this be deployed? My company's min target is iOS17. Can you at least deploy something for the people that are using the latest versions of iOS?

mikenachbaur-okta:
This PR was an initial prototype to determine how to adopt swift concurrency features, but the changes were too large to make all at once.  So this is being broken down into smaller PRs / updates.

PR #220, opened 2025-02-27
Update asynchronous APIs to use Swift Concurrency by default

mikenachbaur-okta:
Traditionally this SDK has implemented all asynchronous operations using completion blocks, GCD, and dispatch queues. This was for maximum compatibility since Swift Concurrency was still fairly new at the time this SDK was first built.  Versions of every one of these APIs also featured `async` implementations that wrapped the completion block-based functions.

In order to take advantage of features such as task cancellation, thread safety and data race prevention, `@Sendable`, and actors, the SDK needs to be updated so the default implementations utilize Swift Concurrency by default, with the completion block-based versions of these functions becoming wrappers that encapsulate the async tasks.

This also has the benefit of eliminating many lines of code and simplifying the implementation.


Sources/AuthFoundation/User Management/Credential.swift
```diff
@@ -182,32 +184,21 @@ public final class Credential: Equatable, OAuth2ClientDelegate {
...
-            switch result {
-            case .success(let token):
-                self.token = token
-                completion(.success(()))
-            case .failure(let error):
-                completion(.failure(error))
-            }
+    public func refresh() async throws {
+        let token = try await oauth2.refresh(token)
+        lock.withLock {
```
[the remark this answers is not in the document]
mikenachbaur-okta: The [follow-up PR improves upon this](https://github.com/okta/okta-mobile-swift/pull/221/files#diff-da8be5bb6f4c028abe02b305d3411a24b59b971884a097464a3d33fc03ce3d13L188-R226), once Swift Actors were integrated in.  The objective here is to ensure that updates to the Credential's `token` property are synchronized across threads so there aren't any concurrent accesses to the same property.  I might backport some of the updates made in the Actors PR though, to make reviewing this one simpler.

Sources/AuthFoundation/User Management/Credential.swift
```diff
@@ -150,7 +150,9 @@ public final class Credential: Equatable, OAuth2ClientDelegate {
     /// The ``UserInfo`` describing this user.
     ///
     /// This value may be nil if the ``userInfo()`` or ``userInfo(completion:)`` methods haven't yet been called.
-    public private(set) var userInfo: UserInfo?
+    public var userInfo: UserInfo? {
```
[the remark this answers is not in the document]
mikenachbaur-okta: This isn't a settable property, but is instead a "computed property".  This is shorthand for:

```
public var userInfo: UserInfo? {
    get {
        userInfoAction.value
    }
}
```

The CoalescedResult object is now used to provide thread-safe access to the previously-fetched value, to eliminate a lot of boilerplate code.

Sources/AuthFoundation/User Management/Internal/CredentialCoordinatorImpl.swift
```diff
@@ -17,23 +17,65 @@ import FoundationNetworking
 #endif
 
 final class CredentialCoordinatorImpl: CredentialCoordinator {
+    private let lock = Lock()
+    private var _credentialDataSource: CredentialDataSource?
     var credentialDataSource: CredentialDataSource {
-        didSet {
-            credentialDataSource.delegate = self
+        get {
+            lock.withLock {
```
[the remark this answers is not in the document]
mikenachbaur-okta: Yes, the lock is used to ensure that only one operation using the same lock can run at a time.  One has to be careful when using this, to avoid calling another function within the lock body that _also_ attempts to use the same lock (aka a "Deadlock"), so I try to use it sparingly.

These properties are seldom set or updated (unless it's in unit tests) but since it's central to the whole system, it's important.  Plus, when you enable Swift 6 mode, allowing for concurrent access to shared mutable state is a compiler error.

Sources/AuthFoundation/Utilities/AsyncUtilities.swift
```diff
@@ -0,0 +1,37 @@
...
+
+/// Convenience function that wraps an expression so that results, or thrown errors, can be propagated to the appropriate underlying delegate functions as necessary.
+/// - Parameters:
+///   - expression: Expression to invoke which should be wrapped by the delegate collection
+///   - success: Closure invoked when the expression is successful
+///   - failure: Closure invoked when the expression throws an error
+/// - Returns: The result when the expression is successful
+@inlinable
+@_documentation(visibility: private)
+public func withExpression<T>(_ expression: () async throws -> T,
```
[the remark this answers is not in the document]
mikenachbaur-okta: I used `withExpression` because I wanted to keep it intentionally vague. It doesn't use or communicate any types related to delegates themselves.  In the next PR I add some other `with*` functions (which is consistent with Apple's functions when dealing with continuations or other lower-level primitives)

Sources/AuthFoundation/Network/APIClient.swift
```diff
@@ -64,7 +64,7 @@ public protocol APIClient {
     func didSend<T>(request: URLRequest, received response: APIResponse<T>)
     
     /// Send the given URLRequest.
-    func send<T: Decodable>(_ request: URLRequest, parsing context: APIParsingContext?, completion: @escaping (Result<APIResponse<T>, APIClientError>) -> Void)
+    func send<T: Decodable>(_ request: URLRequest, parsing context: APIParsingContext?) async throws -> APIResponse<T>
```
stevelind-okta, line 67: Is this a change to an external API? I guess this is then a major version bump?

Sources/AuthFoundation/Network/APIClient.swift
```diff
@@ -361,3 +346,5 @@ let defaultJSONDecoder: JSONDecoder = {
     decoder.keyDecodingStrategy = .convertFromSnakeCase
     return decoder
 }()
+
+var _APIClientRetryDelayTimeIntervalToNanoseconds: Double = 1_000_000_000
```
stevelind-okta, line 350: Why not use one of the Dispatch intervals like `.seconds(1)` to simplify?

Sources/AuthFoundation/Network/APIClientError.swift
```diff
@@ -36,7 +36,7 @@ public enum APIClientError: Error {
     case unsupportedContentType(_ type: APIContentType)
     
     /// Received the given HTTP error from the server.
-    case serverError(_ error: Error)
+    case httpError(_ error: Error)
```
stevelind-okta, line 39: why make this change? usually we get more granular errors than just HTTP codes, for example `EXXXXXXX` https://developer.okta.com/docs/reference/error-codes/

mikenachbaur-okta:
I moved the tests and sample updates to another PR to make the review easier to follow: https://github.com/okta/okta-mobile-swift/pull/223

Sources/AuthFoundation/Network/URLSessionProtocol.swift
```diff
@@ -17,21 +17,32 @@ import FoundationNetworking
...
-public protocol URLSessionDataTaskProtocol {
-    func resume()
-}
-
+@_documentation(visibility: internal)
 extension URLSession: URLSessionProtocol {
-    @_documentation(visibility: internal)
-    public func dataTaskWithRequest(_ request: URLRequest, completionHandler: @escaping (Data?, URLResponse?, Error?) -> Void) -> URLSessionDataTaskProtocol {
-        dataTask(with: request, completionHandler: completionHandler)
+#if os(Linux)
```
stevelind-okta, line 28: Do we tell people we support linux? Unless it's officially supported it would simplify our lives not to do this, who is testing linux?

stevelind-okta (commented):
I haven't been able to do a full review, overall I support the async operations.
Some issues to address:

- Need unit tests for the new functionality
- Some of these items could have been split out, for example `serverError -> httpResponse` as well as the new utility classes for asynchronous + locked operations (with unit tests)
- I'd like to see a spec with the overall arc of these changes and then work breakdown for what is in this particular change

Sources/AuthFoundation/Network/APIClient.swift
mikenachbaur-okta: Yes this is a public API since it's used by the downstream libraries.  It's largely used as internal features of the SDK and while customers aren't expected to use this directly, it still needs to be public since Swift doesn't have any sort of "protected" accessibility option.

And definitely, this will be the SDK's first major release, because of the breaking API changes.

Sources/AuthFoundation/Network/APIClient.swift
mikenachbaur-okta: That requires iOS 16+, so the SDK still uses the older `nanoseconds` notation.

Sources/AuthFoundation/Network/APIClientError.swift
mikenachbaur-okta: There's confusion about the name `serverError` across different error types.  This is for the low-level `APIClient` interface, and represents the error we get back from URLSession.  Elsewhere, the OAuth2Error has a `serverError` type which returns an instance of [OAuth2ServerError](https://okta.github.io/okta-mobile-swift/development/authfoundation/documentation/authfoundation/oauth2servererror) that exposes the code, description, etc returned from the server.

Sources/AuthFoundation/Network/URLSessionProtocol.swift
mikenachbaur-okta: Yes we do. The support is marked as "Experimental", but its been a feature of the SDK since it was first released, and has been a good forcing-function to ensure the API structure is clean & abstract enough to allow developers to extend it as needed without requiring forking.

mikenachbaur-okta:
> * Need unit tests for the new functionality

@stevelind-okta As you and others previously requested, I [split the tests and samples into a separate PR](https://github.com/okta/okta-mobile-swift/pull/223) to make it easier to review.

Sources/AuthFoundation/Utilities/CoalescedResult.swift
```diff
@@ -12,22 +12,106 @@
...
-    
-    func start(_ operation: ((T) -> Void) -> Void) {
-        operation { result in
-            self.finish(result)
+
+    /// Indicates if the asynchronous operation is being performed.
+    nonisolated public var isActive: Bool {
+        let semaphore = DispatchSemaphore(value: 0)
+        nonisolated(unsafe) var result: Bool = false
+        Task {
```
IldarAbdullin-okta, line 45: That pattern looks very risky to me. If property was called from the main thread then Task most likely will be scheduled to asynchronously run in main thread as well. So there is some risk of dead lock here
Probably dispatch_sync or at least Task.detached would be more safe options here

Sources/AuthFoundation/Utilities/CoalescedResult.swift
```diff
@@ -12,22 +12,106 @@
...
         }
+        semaphore.wait()
+        return result
     }
 
-    func finish(_ result: T) {
-        completionHandlers.forEach { completion in
-            completion(result)
+    /// Stores the previously-fetched value, if one has been fetched and its response was successful.
+    nonisolated public var value: T? {
```
IldarAbdullin-okta, line 54: I don't understand what we are doing here and why we need nonisolated property and implement isolation by ourselves(which is risky). We already have isolated property `_value`, why just non return its value in actor's isolated context?

Sources/AuthFoundation/Utilities/CoalescedResult.swift
```diff
@@ -12,22 +12,106 @@
...
+            _value = nil
+        }
+
+        if !_isActive {
+            if let value = _value {
+                return value
+            }
+
+            _isActive = true
+            task = BackgroundTask(named: taskName)
```
IldarAbdullin-okta, line 87: what is the purpose of creating this task variable here?

Sources/AuthFoundation/Utilities/CoalescedResult.swift
```diff
@@ -12,22 +12,106 @@
...
+            willBegin()
```
IldarAbdullin-okta, line 88: Is it fine that begin and end will be called only for the first caller?

Sources/AuthFoundation/Utilities/CoalescedResult.swift
```diff
@@ -12,22 +12,106 @@
...
+            }
         }
+
+        return try await withCheckedThrowingContinuation { continuation in
+            continuations.append(continuation)
+        }
+    }
+
+    private func complete(with result: Result<T, any Error>, didEnd: (Result<T, Error>) -> Void) {
+        _isActive = false
```
IldarAbdullin-okta, line 107: That unblocks reentrance for the errored request. Is it fine? By design?

Sources/AuthFoundation/Utilities/CoalescedResult.swift
```diff
@@ -12,22 +12,106 @@
...
+
+            _isActive = true
+            task = BackgroundTask(named: taskName)
+            willBegin()
+
+            do {
+                let result = try await operation()
+                _value = result
+                complete(with: .success(result), didEnd: didEnd)
+                return result
```
IldarAbdullin-okta, line 94: First caller will get result the very last one. Fine? Would be better to save its continuation similar to other callers?

Sources/OktaDirectAuth/Internal/Extensions/Status+ErrorInitializers.swift
```diff
@@ -0,0 +1,79 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//
+// See the License for the specific language governing permissions and limitations under the License.
+//
+
+import Foundation
+
+extension DirectAuthenticationFlow.Status {
+    @usableFromInline
```
IldarAbdullin-okta, line 16: Why there is so much love to inlinable methods? No concern if sdk is distributed as static lib, but in case of dynamic(let's say linux .so) we are committing ourselves that any change in inlinable function will be a breaking change

IldarAbdullin-okta (commented):
Overall looks good
Posted some questions mainly around actor usage/impl

Sources/AuthFoundation/Utilities/CoalescedResult.swift
alexnachbaur-okta: This is resolved in [the PR where this class becomes an actor](https://github.com/okta/okta-mobile-swift/pull/221), since the `_isActive` async getter will always be invoked on the actor's background thread. It's impossible for an actor's isolation context to be within the main thread unless it is explicitly marked as `@MainActor`.

That being said, you are correct that under very resource-constrained circumstances (where the thread pool has been exhausted). I'll update this in a follow-up PR to use `Task.detached`.

**Note:** It's important to avoid the use of GCD functions wherever possible, so task priority and cancellation can be inherited.

Sources/AuthFoundation/Utilities/CoalescedResult.swift
alexnachbaur-okta: CoalescedResult is used to abstract requests for async resources (e.g. fetching `OpenIdConfiguration`, `UserInfo`, etc).  These values are typically cached at runtime for performance. In the case of OpenIdConfiguration it is requested multiple times throughout a single authentication flow, because there's no guarantee that the application hasn't restarted partway through the workflow.

Lets follow an example of Authorization Code Flow (e.g. web redirect sign in):

1. AuthorizationCodeFlow.start
  1. Fetch OpenIdConfiguration for its authorization endpoint
  2. Compose an authorize URL and pop a browser
  3. Browser returns a redirect URI
2. AuthorizationCodeFlow.resume
  1. Extract the authorization code
  2. Fetch OpenIdConfiguration for the token endpoint
  3. Request a token from the server
3. Validate the server's token response (run in parallel with the 2.3. token request)
  1. Fetch the OpenIdConfiguration for the keys endpoint
  2. Fetch the jwks keys
  3. Validate the token

This workflow uses the async versions of these functions, but 3rd-party developers use these resources for their own needs as well, and often from non-async contexts.  For this reason, the OAuth2Client class has a property which returns the previously-fetched [openIdConfiguration value](https://okta.github.io/okta-mobile-swift/development/authfoundation/documentation/authfoundation/oauth2client/openidconfiguration).

For that reason, the CoalescedResult provides nonisolated access to prevent repetitive boilerplate throughout the other uses in the SDK.

Sources/AuthFoundation/Utilities/CoalescedResult.swift
alexnachbaur-okta: `BackgroundTask` is a cross-platform abstraction for UIBackgroundTaskIdentifier and UIApplication.beginBackgroundTask.

An important OAuth2 security feature is rotating refresh tokens, whereby refresh tokens are single-use. Using a refresh token in this case returns a new `access_token` _and_ `refresh_token` in the response, which the client needs to save for subsequent refreshes. If an app has a stale refresh token, the user will be required to sign in again once the new `access_token` expires.

This situation can happen all too easily if the user backgrounds the app while a token is being refreshed.  This is because the refresh operation begins on the server, but the client app's process is suspended before it can receive the response, meaning it's unable to store the new refresh token in the keychain.

To that end, on platforms that support it, we need to ensure the network responses fully finish in the background before the app is suspended.

To prevent a lot of unnecessary boilerplate, the CoalescedResult type retains a reference to BackgroundTask, which encapsulates the details for how this is implemented per-platform (currently only on UIKit-based platforms, but could be expanded in the future). This way, all network operations run in this manner will keep the app alive while backgrounded until their continuations have had a chance to complete (e.g. tokens have been stored).

Sources/AuthFoundation/Utilities/CoalescedResult.swift
alexnachbaur-okta: Yes, these should only be called once, regardless of the number of callers invoking it.  This is typically because `willBegin` and `didEnd` trigger calls to delegates, and duplicated notifications can cause problems for delegate consumers.

Sources/AuthFoundation/Utilities/CoalescedResult.swift
alexnachbaur-okta: Yes that's correct. The CoalescedResult should only allow a single live operation at a time, coalescing multiple concurrent calls into one actual network request (or some other async operation).  But once it's complete, and all callers have been notified, a subsequent request can be made later.

Since this type is an actor, it doesn't matter that the `_isActive` is set first, since any subsequent call to `perform` will be serialized.

Typically the CoalescedResult values are stored as `let` values to limit the number of mutable values in the library, which simplifies `Sendable` conformance.

Sources/AuthFoundation/Utilities/CoalescedResult.swift
alexnachbaur-okta: Yes this is fine. There's no guarantee that the callers of the coalesced operation get their result value in the order they were called. But I'll make sure to document this in the API docs to make this explicit, so there's no misunderstandings in the future.

Sources/AuthFoundation/OAuth2/OAuth2Client.swift
```diff
@@ -238,332 +159,134 @@ public final class OAuth2Client {
...
-    internal var openIdConfigurationAction: CoalescedResult<Result<OpenIdConfiguration, OAuth2Error>>?
-
-    private lazy var jwksQueue: DispatchQueue = {
-        DispatchQueue(label: "com.okta.jwksQueue.\(baseURL.host ?? "unknown")",
-                      qos: .userInitiated,
-                      attributes: .concurrent)
-    }()
-    internal var jwksAction: CoalescedResult<Result<JWKS, OAuth2Error>>?
+    let openIdConfigurationAction = CoalescedResult<OpenIdConfiguration>(taskName: "OpenIdConfiguration")
+    let jwksAction = CoalescedResult<JWKS>(taskName: "OpenIdConfiguration")
```
alexnachbaur-okta, line 274: Whoops, I need to change this task name

Package.swift
```diff
@@ -7,8 +7,8 @@ var package = Package(
     name: "AuthFoundation",
     defaultLocalization: "en",
     platforms: [
-        .iOS(.v12),
-        .tvOS(.v12),
+        .iOS(.v13),
```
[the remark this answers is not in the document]
alexnachbaur-okta: @umangshah-okta We have [a documented support policy in the README](https://github.com/okta/okta-mobile-swift?tab=readme-ov-file#support-policy) (which will need to be updated in another PR when we craft the release update), but basically we support the latest 4 major iOS versions, but we support older versions in a "best-effort" capacity.

If we drop support for a version older than 4 releases prior (e.g. because it's impractical or impossible to do so), we don't consider it a breaking API change and can do that in a minor release.

However, since this update will constitute a major version release there are no such limitations.  Regardless, technically speaking, we only need to support iOS 14+ to meet our current support policy, so iOS 13 is still a "best effort" supported version.

Sources/AuthFoundation/Network/APIClient.swift
```diff
@@ -163,42 +163,34 @@ extension APIClient {
     
     public func didSend<T>(request: URLRequest, received response: APIResponse<T>) {}
     
-    public func send<T>(_ request: URLRequest, parsing context: APIParsingContext? = nil, completion: @escaping (Result<APIResponse<T>, APIClientError>) -> Void) {
```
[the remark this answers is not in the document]
alexnachbaur-okta: The APIClient interfaces aren't heavily used by customers, and largely are public because they're needed by other libraries within the SDK.

Under normal circumstances I'd provide completion-block deprecation features, but since this is a major update, I feel it's acceptable to "rip off the bandaid" and lean into the Swift Concurrency approach now.

**Note:** This interface will be used within a subsequent release when we add support for DPoP and an automatically-authorized Resource Server client.

Sources/AuthFoundation/Network/URLSessionProtocol.swift
```diff
@@ -17,21 +17,32 @@ import FoundationNetworking
...
+    public func data(for request: URLRequest) async throws -> (Data, URLResponse) {
```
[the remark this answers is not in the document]
alexnachbaur-okta: There's a known issue where the URLSession `async` functions haven't yet been ported over to the Swift Foundation library, so they're not yet available in Linux.  You might notice in other parts of the SDK where we include the following import within any class that references URLSession or URLRequest types

```swift
#if os(Linux)
import FoundationNetworking
#endif
```

Linux has the network-related portions of Foundation in a different target library, and not everything has been updated yet.

Sources/AuthFoundation/OAuth2/OAuth2Client.swift
```diff
@@ -45,12 +45,16 @@ public final class OAuth2Client {
...
+    public var openIdConfiguration: OpenIdConfiguration? {
+        openIdConfigurationAction.value
+    }
 
     /// The ``JWKS`` key set for this org.
     ///
     /// This value will be `nil` until the keys have been retrieved through the ``jwks(completion:)`` or ``jwks()`` functions.
-    public private(set) var jwks: JWKS?
+    public var jwks: JWKS? {
+        jwksAction.value
```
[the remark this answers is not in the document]
alexnachbaur-okta: We discussed this on Zoom, but this is using the nonisolated value exposed by the `CoalescedResult` type.

Sources/AuthFoundation/OAuth2/OAuth2Client.swift
alexnachbaur-okta: I fixed this [in the follow-up PR](https://github.com/okta/okta-mobile-swift/pull/228).

umangshah-okta (approved):
looks good 🚢

Package.swift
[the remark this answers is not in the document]
umangshah-okta: got it

Sources/AuthFoundation/Network/APIClient.swift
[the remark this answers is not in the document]
umangshah-okta: got it, makes sense

Sources/AuthFoundation/Network/APIRequest.swift
```diff
@@ -244,24 +236,9 @@ extension APIRequest {
...
-    }
-    
-    @available(iOS 13.0, tvOS 13.0, macOS 10.15, watchOS 6, *)
     public func send(to client: APIClient, parsing context: APIParsingContext? = nil) async throws -> APIResponse<ResponseType> {
-        try await withCheckedThrowingContinuation { continuation in
-            send(to: client, parsing: context) { result in
-                continuation.resume(with: result)
-            }
-        }
+        try await client.send(try request(for: client),
```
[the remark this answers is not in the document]
umangshah-okta: got it

Sources/AuthFoundation/Network/URLSessionProtocol.swift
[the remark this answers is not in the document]
umangshah-okta: got it

Sources/AuthFoundation/OAuth2/OAuth2Client.swift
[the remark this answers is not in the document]
umangshah-okta: got it

PR #221, opened 2025-03-11
Adopt actors and locks to avoid data races and thread consistency

alexnachbaur-okta:
This update builds upon the previous PR #220, leveraging more Swift Concurrency features to avoid data races, ensure consistency when critical operations (such as token storage) are performed, while maintaining an easy-to-use developer experience.

The primary changes here are:
* Create a `CredentialActor` global actor which the Token Storage and CredentialCoordinator subsystems use to ensure consistency and multithreaded data safety when these foundational systems are used.
* Update all `AuthenticationFlow` classes to be actors, simplifying data safety for concurrent operations.
* Convert `Token` to be a `struct` to officially make it an immutable object.
* Introduce the use of Locks or DispatchSemaphores, where appropriate, to allow frequently-used objects (e.g. Credential, OAuth2Client, etc) to safely expose synchronous functions or properties that interact with actors. This includes conveniences (e.g. `withAsyncGroup` or `withAsyncThrowingGroup`) for consistency throughout the SDK.
* Update various types to conform to `Sendable`, allowing them to be used with Actors
* Removed the last remaining usage of detached blocks, opting for Tasks, improving support for task cancellation

**Note:** To simplify the review, the tests and sample app changes were moved to a separate PR #222


mikenachbaur-okta:
**Note:** I'm making some minor updates to improve the developer experience when using AuthenticationFlow actors, and to clean up the unit tests.

Sources/AuthFoundation/Debugging/APIRequestObserver.swift
```diff
@@ -53,26 +53,46 @@ public final class DebugAPIRequestObserver: OAuth2ClientDelegate {
...
+                        forName: .oauth2ClientCreated,
+                        object: nil,
+                        queue: nil,
+                        using: { [weak self] notification in
+                            guard let self = self,
+                                  let client = notification.object as? OAuth2Client
+                            else {
+                                return
+                            }
+                            client.add(delegate: self)
```
IldarAbdullin-okta, line 91: Does this require synchronization? Or client cares about synchronization under the hood?

Sources/AuthFoundation/OAuth2/OAuth2Client.swift
```diff
@@ -32,16 +32,27 @@ extension OAuth2ClientDelegate {
...
     /// The URLSession used by this client for network requests.
     public let session: URLSessionProtocol
     
     /// The configuration that identifies this OAuth2 client.
     public let configuration: Configuration
     
     /// Additional HTTP headers to include in outgoing network requests.
-    public var additionalHttpHeaders: [String: String]?
-    
+    nonisolated(unsafe) public var additionalHttpHeaders: [String: String]? {
```
IldarAbdullin-okta, line 43: What `nonisolated` means for the class?

Sources/AuthFoundation/OAuth2/OAuth2Client.swift
```diff
@@ -96,8 +107,11 @@ public final class OAuth2Client {
         
         self.configuration = configuration
         self.session = session ?? URLSession(configuration: .ephemeral)
-        
-        NotificationCenter.default.post(name: .oauth2ClientCreated, object: self)
+        self.refreshQueue = DispatchQueue(label: "com.okta.refreshQueue.\(configuration.issuerURL.host ?? "unknown")",
+                                          qos: .userInitiated,
+                                          attributes: .concurrent)
+
+        TaskData.notificationCenter.post(name: .oauth2ClientCreated, object: self)
```
IldarAbdullin-okta, line 114: That pattern has some pitfalls as we send notification with `self` for a partially constructed/initialized object. Since notification center calls all callbacks synchronously it means that all callbacks will be executed within `init` context. There is no guarantee that observers will not start using `self` immediately(which is still being constructed)

Sources/AuthFoundation/User Management/Credential.swift
```diff
@@ -19,18 +19,30 @@ import FoundationNetworking
...
-public final class Credential: Equatable, OAuth2ClientDelegate {
+public final class Credential: Sendable, Equatable, OAuth2ClientDelegate {
     /// The current or "default" credential.
     ///
     /// This can be used as a convenience to store a user's token within storage, and to access the user in a safe way. If the user's token isn't stored, this will automatically store the token for later use.
     public static var `default`: Credential? {
-        get { coordinator.default }
-        set { coordinator.default = newValue }
+        get {
+            withAsyncGroup { @CredentialActor in
```
IldarAbdullin-okta, line 28: Why in some cases we use Lock objects and in other cases are calling `withAsyncGroup`?

Sources/AuthFoundation/User Management/Credential.swift
```diff
@@ -111,15 +135,15 @@ public final class Credential: Equatable, OAuth2ClientDelegate {
...
+    public var id: String { token.id }
     
     /// The metadata associated with this credential.
     ///
     /// This property can be used to associate application-specific information with a ``Token``. This can be used to identify which token should be associated with certain parts of your application.
     ///
     /// > Important: Errors thrown from the setter are silently ignored. If you would like to handle errors when changing metadata, see the ``setTags(_:)`` function.
     public var tags: [String: String] {
-        get { _metadata.tags }
+        get { metadata.tags }
```
IldarAbdullin-okta, line 146: withAsyncGroup?

Sources/AuthFoundation/User Management/Credential.swift
```diff
@@ -157,13 +186,20 @@ public final class Credential: Equatable, OAuth2ClientDelegate {
...
+        get {
+            lock.withLock { _automaticRefresh }
+        }
+        set {
+            lock.withLock {
+                guard _automaticRefresh != newValue else { return }
+                _automaticRefresh = newValue
+
+                if _automaticRefresh {
+                    startAutomaticRefresh()
```
IldarAbdullin-okta, line 199: How long this can execute? Note, the block still acquires lock. Is it that necessary to keep property locked until function is not finished?

Sources/AuthFoundation/User Management/Credential.swift
```diff
@@ -301,55 +337,127 @@ public final class Credential: Equatable, OAuth2ClientDelegate {
...
-        guard let timerSource = createAutomaticRefreshTimer() else { return }
+        guard let expiresAt = _token.expiresAt
+        else {
+            return
+        }
+        let graceInterval = Credential.refreshGraceInterval
+        let timeOffset = max(0.0, expiresAt.timeIntervalSinceNow - Date.nowCoordinated.timeIntervalSinceNow - graceInterval)
+        let repeatInterval = min(_token.expiresIn - graceInterval, _token.expiresIn)
+
+        _automaticRefreshTask = Task(priority: .medium) {
```
IldarAbdullin-okta, line 397: What if task is already created and active?

Sources/AuthFoundation/User Management/Credential.swift
```diff
@@ -301,55 +337,127 @@ public final class Credential: Equatable, OAuth2ClientDelegate {
...
+            } while !Task.isCancelled
+        }
     }
-    
-    func stopAutomaticRefresh() {
-        automaticRefreshTimer?.cancel()
-        automaticRefreshTimer = nil
+
+    private func stopAutomaticRefresh() {
+        _automaticRefreshTask?.cancel()
```
IldarAbdullin-okta, line 427: There is no synchronization with `startAutomaticRefresh` function. `_automaticRefreshTask ` can be overwritten with new task, so previous will never be cancelled

Sources/AuthFoundation/User Management/Internal/CredentialCoordinatorImpl.swift
```diff
@@ -165,28 +187,31 @@ final class CredentialCoordinatorImpl: CredentialCoordinator {
...
 
-    func observe(oauth2 client: OAuth2Client) {
+    nonisolated func observe(oauth2 client: OAuth2Client) {
         client.add(delegate: self)
     }
 }
 
 extension CredentialCoordinatorImpl: OAuth2ClientDelegate {
-    func oauth(client: OAuth2Client, didRefresh token: Token, replacedWith newToken: Token?) {
+    nonisolated func oauth(client: OAuth2Client, didRefresh token: Token, replacedWith newToken: Token?) {
```
IldarAbdullin-okta, line 202: Why nonisolated if `withAsyncGroup` guarantees isolation?

Sources/AuthFoundation/Utilities/AsyncUtilities.swift
```diff
@@ -12,6 +12,34 @@
 
 import Foundation
 
+/// Shared TaskLocal data used within AuthFoundation tasks.
+///
+/// This is primarily used by unit tests to ensure the consistent passage of shared framework objects to isolate tests.
+@_documentation(visibility: internal)
+public enum TaskData {
+    /// The NotificationCenter instance that should be used when posting or observing notifications.
+    @TaskLocal public static var notificationCenter: NotificationCenter = .default
```
IldarAbdullin-okta, line 21: What is the intention of @TaskLocal here?

IldarAbdullin-okta (commented):
Overall looks good. Posted some comments, please review

Sources/AuthFoundation/Debugging/APIRequestObserver.swift
alexnachbaur-okta: This function is introduced by [the UsesDelegateCollection protocol](https://okta.github.io/okta-mobile-swift/development/authfoundation/documentation/authfoundation/usesdelegatecollection).  Since the `delegateCollection` property is a `let` value, and is `Sendable`, interacting with it from the OAuth2Client is considered safe.

I do notice that the DelegateCollection type itself is potentially susceptible to race conditions with mutations to its underlying `delegates` value, so I'll include an update that synchronizes concurrent access that mutates the list of delegates.

Sources/AuthFoundation/OAuth2/OAuth2Client.swift
alexnachbaur-okta: `nonisolated` means a property can be used from any isolation context (e.g. any thread, actor, etc).  The `unsafe` attribute is included because the property is mutable, and that we're using our own thread safety checks outside of Swift Concurrency's built-in features (e.g. in this case, a `os_unfair_lock`).

Sources/AuthFoundation/OAuth2/OAuth2Client.swift
alexnachbaur-okta: The initializer has already completed at this point, and is wiring up its communication with the credential coordinator.  Fortunately the compiler won't let `self` be used until all its properties have been initialized, so at this point it's safe for notification observers to use the object.

The reason why it's important is that there are many occasions where an OAuth2Client may be implicitly created on behalf of another actor in the system (e.g. an AuthenticationFlow), and it could be important for a delegate to observe events / update the object before it's used.

Sources/AuthFoundation/User Management/Credential.swift
alexnachbaur-okta: This function is renamed in a subsequent PR to `withIsolationSync` and `withIsolationSyncThrowing`, which more accurately describes what's going on.

The goal isn't to arbitrarily lock a process or a thread, but rather to interact with a property or synchronous function within an actor in a different isolation context, which forces the call to be `async`.  When performing this access from a `nonisolated` property or function, we still need to ensure that the call to the underlying actor is made within the appropriate isolation context, while ensuring task cancellation and priority inheritance is respected.

Since we do this in a number of places, I abstracted this logic out into a reusable `inlinable` function.

Sources/AuthFoundation/User Management/Credential.swift
```diff
@@ -19,18 +19,30 @@ import FoundationNetworking
 /// Convenience object that provides methods and properties for using a user's authentication tokens.
 ///
 /// Once a user is authenticated within an application, the tokens' lifecycle must be managed to ensure it is properly refreshed as needed, is stored in a secure manner, and can be used to perform requests on behalf of the user. This class provides capabilities to accomplish all these tasks, while ensuring a convenient developer experience.
-public final class Credential: Equatable, OAuth2ClientDelegate {
+public final class Credential: Sendable, Equatable, OAuth2ClientDelegate {
     /// The current or "default" credential.
     ///
     /// This can be used as a convenience to store a user's token within storage, and to access the user in a safe way. If the user's token isn't stored, this will automatically store the token for later use.
     public static var `default`: Credential? {
```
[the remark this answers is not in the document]
alexnachbaur-okta: The `Credential` class is the primary entry point developers use for interacting with tokens, storage, and conveniences.  If we make this class a part of `@CredentialActor`, every access to it would need to be `async`, which would wreck the developer experience.

The `CredentialActor` global actor is really a part of [the Token Management subsystem](https://okta-client-design.netlify.app/design/SDKGuidelines/TokenManagement/) of the SDK.  If anything, I should consider renaming this to the `@TokenManagement` global actor.

Sources/AuthFoundation/User Management/Credential.swift
alexnachbaur-okta: The `metadata` property has custom getter/setter which uses the credential's local `lock` to synchronize access to the `nonisolated(unsafe) _metadata` property. From there, the Metadata type is `Sendable`, and its `tags` property is a `let` value, so it's safe to access across isolation contexts.

Sources/AuthFoundation/User Management/Credential.swift
alexnachbaur-okta: This function is improved upon within a subsequent PR, but ultimately starting and stopping the timer is a very short operation, since the body of the function almost immediately dispatches to a cancellable task.

Sources/AuthFoundation/User Management/Credential.swift
alexnachbaur-okta: `start` and `stopAutomaticRefresh` are only invoked from the `automaticRefresh` property's setter.  But even so, in the event a future code change is made which breaks this, once the old Task's cancellation handler is overwritten, it will lose its retention count and will be cancelled & destroyed.

Sources/AuthFoundation/User Management/Credential.swift
alexnachbaur-okta: These functions are private, and are only called from the `automaticRefresh` setter.  They're locked at the setter level, so special care is taken within these functions to ensure that no inadvertent accesses are made to public nonisolated properties which use the locks (e.g. it uses `_token` instead of `token`).

Additionally, when the task needs to be cancelled in response to an async operation within the Task itself, it re-locks its access to `self` before niling out the refresh task.

Sources/AuthFoundation/User Management/Internal/CredentialCoordinatorImpl.swift
alexnachbaur-okta: It needs to be marked as `nonisolated` because the OAuth2ClientDelegate protocol is not marked as `Sendable`, so we need to explicitly assign `nonisolated` to the function declaration since CredentialCoordinatorImpl _is_ `Sendable` (due to its conformance to `CredentialCoordinator`).

Sources/AuthFoundation/Utilities/AsyncUtilities.swift
alexnachbaur-okta: https://developer.apple.com/documentation/swift/tasklocal

It's part of Swift Concurrency that allows for contextual state to be inherited across a series of related tasks.  It's the foundation for SwiftUI's `environment` and `environmentObject` features.

Instead of using NotificationCenter's singleton when posting messages, this allows the SDK, within a task, to share state across its various calls.  This is used within unit tests to ensure that parallel tests don't clobber each other, causing false-negatives.  During regular operation, and unless explicitly overridden within a Task, its value will be the `NotificationCenter.default` singleton.

Sources/AuthFoundation/OAuth2/OAuth2Client.swift
IldarAbdullin-okta: It still looks weird to me that we postpone construction of the object with blocking notification. That means that `let obj = Object()` call will be postponed until all subscribers finish their processing(and we have no idea how heavy it can be)
I suggest either fire this notification in async task or use NotificationQueue instead

alexnachbaur-okta: I feel like dispatching this async is the correct approach, but I'd need better testing there to ensure the appropriate consumers get notified in a timely fashion.  Or, even better, the internal consumers of this notification that need guaranteed access to the object before it's used downstream, should have a private way of being notified synchronously.

I'll look into this ...

PR #224, opened 2025-03-15
Enable Swift 6 support with strict concurrency checking

alexnachbaur-okta:
This introduces several important changes:

* Separates Swift 5.10 support into a separate `Package@swift-5.10.swift` file to enable Swift 6 PackageDescription syntax.
* Runs matrix tests across Xcode 15.4 and 16.2, and Swift 5.10 and 6.0
* Uses the new DocC combined documentation experimental feature to enable linking of related references between targets
* Enable opt-in Swift 6 features into Swift 5.10, including:
  * ExistentialAny
  * ForwardTrailingClosures
  * InferSendableFromCaptures
  * IsolatedDefaultValues
  * DisableOutwardActorInference
  * GlobalConcurrency
  * RegionBasedIsolation
* Enable Strict Concurrency Checking across all library targets in both Swift 6 and 5.10
* Improve actor context and isolation boundaries when using async convenience wrappers
* Marked common properties on `AuthenticationFlow` actors (`isAuthenticating` / `context`) as `nonisolated`

> **Note:** Linux only supports Swift 6 since Concurrency features are wildly broken on 5.10.


IldarAbdullin-okta (approved):
LGTM!

Why there is so much love to poorly documented features like `@_inheritActorContext`?

alexnachbaur-okta:
@IldarAbdullin-okta While the `@_inheritActorContext` attribute _is_ an underscore attribute, it still provides a crucial hint to the compiler that the given block should inherit the context from its actor. This is used within the `Task.init` initializer, as well as a few other places within the Swift Foundation codebase.

Here's a relevant quote [from a post on the Swift developer forums](https://developer.apple.com/forums/thread/761150):
> This tells the compiler that the closure argument should inherit its isolation from the surrounding code. 

By using this attribute, we ensure the caller can execute the function, passing a closure, ensuring that whatever actor is being used by the caller is propagated to the supplied block.  This eliminates _a ton_ of thread synchronization and multithreading risks.  You'll notice that the `withExpression` function in this PR doesn't mark any of those closures as `@escaping`, because they're meant to be run within the same local context.  But without `@_inheritActorContext`, the compiler won't know to pass the Task isolation context along to the closure.

Issue #225, opened 2025-03-15
Swift 6 Support


alexnachbaur-okta:
Swift has been incrementally introducing new concurrency features to the language which protects developers from a variety of hard-to-diagnose runtime errors. Many of these features were only available if a developer opted-in to those features.  With the release of Swift 6, these optional features are enabled by default, which has a huge impact on developers.

However, these language features has a significant impact on the ability to product applications from a variety of problems, such as data races, preventing concurrent access to shared mutable state, and many other tricky edge-cases.

The Okta Client SDK has been stable for a long time, but recent updates to Swift and Apple's platforms caused rare threading issues to become much more common, negatively impacting users and introducing tricky race conditions. This is most prevalent in the [race conditions around the default credential property](https://github.com/okta/okta-mobile-swift/issues/219).

To that end, the Okta Client SDK is being updated to fully support Swift 6 concurrency features, will harden its underlying foundations to leverage modern language capabilities, and to eliminate entire classes of runtime multithreading safety issues.

### Planned Changes

To address this, the following changes are planned:

* Update `AuthenticationFlow` to more consistently handle mutable state by adopting the concept of a `Context` typealias which will be used to store in-progress and developer-assigned state.
* Adapt functions to use `async` / `await` by default with completion-block convenience wrappers, rather than the other way around.
* Update foundational classes to be `actor` types to simplify concurrent access, while exposing synchronous APIs to developers for convenient access (e.g. no leaking async operations across your codebase), leveraging custom global actors for the token storage / credential subsystem.
* Enable strict concurrency checking and other language features, annotating the APIs with the appropriate `@Sendable` and other types, and add complete Swift 6 language support, with backwards compatibility to Swift 5.10.

### Pull Requests

This undertaking is massive, and impacts every file within the SDK. This work is therefore being broken down into separate pull requests to simplify review, and to reduce risk.

The following PRs, in order, constitute this work:

1. #214 
2. #217 
3. #220 
4. #223 
5. #221 
6. #222 
7. #224 
8. #228 

### Review & Feedback

I encourage developers to review these PRs as much as you like, and to use this issue to ask questions, provide feedback, and to stay informed about updates to this very important issue.

AlexNachbaur:
The [Okta Client SDK version 2.0.0](https://github.com/okta/okta-mobile-swift/releases/tag/2.0.0) was released which resolves this, and other issues.  Thank you everyone for your patience!

PR #229, opened 2025-05-06
Ensure Keychain SecItemDelete is not called with unsupported attributes

alexnachbaur-okta:
Under some circumstances, requests to delete a Keychain item could result in calls that used invalid request attributes / filters, or includes too many arguments that prevents old values from being deleted.

### DIFF
--- a/Sources/AuthFoundation/Security/Internal/Keychain+Extensions.swift
+++ b/Sources/AuthFoundation/Security/Internal/Keychain+Extensions.swift
@@ -47,6 +47,16 @@ fileprivate let _invalidDeleteQueryKeys = [
     kSecReturnPersistentRef,
 ].map { $0 as String }
 
+extension Keychain {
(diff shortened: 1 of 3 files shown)


IldarAbdullin-okta (approved):
LGTM!

pxmal:
Hi @alexnachbaur-okta 

When do you expect to have 2.0.0 out? I'm afraid some of our users may be affected by this issue

Edit: I'm only asking because in the README the stable release is 2.0.0 so I assumed that it may be soon
Edit2: And I'm using SPM

AlexNachbaur:
@pxmal The primary work itself is complete, but since this will be the SDK's first major breaking change, there are other release & update tasks necessary, both for the [okta-idx-swift library](https://github.com/okta/okta-idx-swift), as well as other version migration & documentation updates.

[Please watch this issue for updates](https://github.com/okta/okta-mobile-swift/issues/225).

PR #230, opened 2025-05-27
Updates to ensure OktaIdx supports the Swift6 updates

alexnachbaur-okta:
* Streamline error handling
* Improve APIRequestPollingHandler to be more extensible
* Centralize redirect URI code parsing and validation
* Update JSON types to simplify input value handling


Sources/AuthFoundation/Utilities/CoalescedResult.swift
```diff
@@ -56,7 +56,7 @@ public actor CoalescedResult<T: Sendable>: Sendable {
     nonisolated public var value: T? {
         let semaphore = DispatchSemaphore(value: 0)
         nonisolated(unsafe) var result: T?
-        Task.detached {
+        Task.detached(priority: Task.currentPriority) {
```
alexnachbaur-okta, line 59: Note: This change was introduced due to a runtime warning by Xcode, indicating that a user-initiated operation resulted in a child task which had a lower priority than the parent task.

These are the sorts of benefits we get from adopting Swift concurrency, and task inheritance! 🎉

FeiChen-okta (approved):

PR #262, opened 2025-11-19
Merge release branch 2.1.3 into master

AlexNachbaur:


Sources/BrowserSignin/BrowserSignin.swift
jakubKozakW: Hello there! 
Since this is a public enum, shouldn't this be considered as a breaking change and thus require a major version bump? 🤔 
It breaks switch statements and case matching in an existing code at least.
Thanks in advance!

PR #269, opened 2026-01-22
Add support for Pushed Authorization Requests

AlexNachbaur:
Introduce conditional support for Pushed Authorization Requests during the web redirect sign-in flow. If the PAR request is unable to compose an authorization URL it falls back to a traditional query-string approach.

### DIFF
--- a/Sources/OAuth2Auth/Authentication/AuthorizationCodeFlow+Context.swift
+++ b/Sources/OAuth2Auth/Authentication/AuthorizationCodeFlow+Context.swift
@@ -94,7 +94,10 @@ extension AuthorizationCodeFlow {
                 authenticationURL = nil
             }
         }
-
+        
+        /// Utilize Pushed Authorization Requests (PAR) if supported by the Authorization Server.
+        public var pushedAuthorizationRequestEnabled: Bool = true
+        
         /// The current authentication URL, or `nil` if one has not yet been generated.
         public internal(set) var authenticationURL: URL?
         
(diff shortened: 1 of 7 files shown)


Sources/OAuth2Auth/Authentication/AuthorizationCodeFlow+Context.swift
```diff
@@ -94,7 +94,10 @@ extension AuthorizationCodeFlow {
                 authenticationURL = nil
             }
         }
-
+        
+        /// Utilize Pushed Authorization Requests (PAR) if supported by the Authorization Server.
+        public var pushedAuthorizationRequestEnabled: Bool = true
```
FeiChen-okta, line 99: should this default to false?

FeiChen-okta (approved):

Sources/OAuth2Auth/Authentication/AuthorizationCodeFlow+Context.swift
AlexNachbaur: In general PAR is preferred to the standard "Include everything in one giant query string" approach, so defaulting to `true` is the desired behavior.  Explicitly setting it to false is mainly an escape hatch in the event that somebody wants to do something crazy with it, or if they know they're authenticating against an authorization server that has broken PAR support.

This update checks the `.well-known/openid-configuration` endpoint to identify if PAR is available before attempting to use it anyway, so query-string-encoded authentication is used as a fallback even when this property is set to `true`.

FeiChen-okta (approved):

PR #274, opened 2026-03-31
Adapt CredentialActor to use a GCD-based serial queue to avoid mixing thread isolation patterns

AlexNachbaur:
* Updates SDKVersion User-Agent processing to avoid using async UIDevice/WKInterfaceDevice functions.
* Introduce a GCD-based serial queue to the CredentialActor to avoid mixing task isolation implementations vis-á-vis  withIsolationSync. This fixes problems with thread starvation in busy applications or with low-core systems.
* Add tests to cover thread starvation to validate the fixes

### DIFF
--- a/Sources/AuthFoundation/Utilities/AsyncUtilities.swift
+++ b/Sources/AuthFoundation/Utilities/AsyncUtilities.swift
@@ -35,16 +35,93 @@ extension Task where Success == Never, Failure == Never {
     }
 }
 
+private final class DispatchQueueExecutor: SerialExecutor {
+     private let queue: DispatchQueue
+
+     init(queue: DispatchQueue) {
+         self.queue = queue
+     }
+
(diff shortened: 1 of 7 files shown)


Sources/AuthFoundation/Utilities/AsyncUtilities.swift
```diff
@@ -35,16 +35,93 @@ extension Task where Success == Never, Failure == Never {
     }
 }
 
+private final class DispatchQueueExecutor: SerialExecutor {
+     private let queue: DispatchQueue
+
+     init(queue: DispatchQueue) {
```
IldarAbdullin-okta: How we guarantee that is a serial queue passed and not a concurrent one?

IldarAbdullin-okta (approved):
LGTM!

FeiChen-okta (approved):

PR #277, opened 2026-05-19
Fix #254 — User-Agent header missing in Release builds

### LINKED ISSUE #254: User-Agent header missing in Release builds
### Describe the bug?

When using the okta-mobile-swift SDK, network requests made from Release builds do not include the User-Agent header.

In Debug builds, the User-Agent header is included as expected.

All Okta SDK network requests should include a User-Agent header in both Debug and Release configurations, for example:

User-Agent: okta-authfoundation-swift/2.1.1 iOS/18.1 Device/arm64

Debug builds:
User-Agent header is present ✅

Release builds:
User-Agent header is missing ❌

### Reproduction Steps?

Run the SDK login flow in a Debug build → capture request headers → User-Agent is present.

Run the same flow in a Release/TestFlight build → capture request headers → User-Agent is missing.

### Additional Information?

_No response_

### SDK Version(s)

2.1.1

### Build Information

_No response_

RicardoKoch:
SDKVersion.userAgent is populated as a side effect of evaluating the lazy 'static let' registration in each Version+*.swift. Every dereference site is currently spelled 'assert(SDKVersion.X != nil)' — but Swift's assert(_:) takes its condition as @autoclosure, which the optimizer strips in -O builds. In Release builds the lazies never fire, SDKVersion._userAgent stays the empty string, and every outgoing request goes out with an empty User-Agent header.

Replace the 19 'assert(SDKVersion.X != nil)' triggers across 13 files with '_ = SDKVersion.X' so the lazy registration fires regardless of build configuration. Same registration mechanism — only the trigger moves out of an @autoclosure.

Adds Tests/AuthFoundationTests/SDKVersionRegistrationTests.swift to pin the contract. Verified to fail in 'swift test -c release' before this change and pass after.

### DIFF
--- a/Sources/AuthFoundation/Migration/Migration.swift
+++ b/Sources/AuthFoundation/Migration/Migration.swift
@@ -43,7 +43,7 @@ public final class Migration {
     nonisolated(unsafe) private(set) var registeredMigrators: [any SDKVersionMigrator]
 
     init(migrators: [any SDKVersionMigrator]? = nil) {
-        assert(SDKVersion.authFoundation != nil)
+        _ = SDKVersion.authFoundation
         self.registeredMigrators = migrators ?? Self.defaultMigrators
     }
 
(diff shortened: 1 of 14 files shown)


AlexNachbaur:
Thank you for this @RicardoKoch, I have pulled this commit into #283 and will close this issue.

